PR #3, opened 2020-04-04
Athena/components

ath-chen:
- made /story up to latest specs 
- work to be done: 
   - need to pass in proper props from ajax calls --> owner panel's progress bar / just put variables there 
      for now as a template 
   - create photos component & need to pass in


src/components/HeroBanner/styles.module.scss
```diff
@@ -1,7 +1,29 @@
 
-.image {
-  object-fit: cover;
+.container {
   width: 100%;
-  max-height: 511px;
-  min-height: 250px;
+  border-bottom: 2px solid #eeeeee;
+  padding: 25px 50px 22px;
```
jtmckibb, line 5: Interesting that these aren't multiples of 8

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -9,16 +9,69 @@ interface Props {
...
+  
+  // these will be props fix throughout when passed thru 
+  const current = 4000;
+  const need = 5000;
+  //  fix ^^^ 
+
+  
   return (
     <section className={classnames(styles.container, className)}>
+      <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css" />
```
jtmckibb, line 21: What's this?

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -9,16 +9,69 @@ interface Props {
...
+        {/* pass props down here*/}
+        <div>${current} of ${need}</div>
+      </div>
+
       <div className={styles.buttonContainer}>
         <button className={classnames(styles.button, "button--filled")}>Donate</button>
         <button className={classnames(styles.button, "button--outlined")}>Buy Gift Card</button>
       </div>
+
+      <div className={styles.summaryContainer}>
```
jtmckibb, line 42: Not relevant to this commit, but this section is going to be hidden for MVP. When you get info from the endpoint, if this info doesn't exist, we should hide it.

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -9,16 +9,69 @@ interface Props {
...
+          <span>3 months</span>
+        </div> <br/>
+        <div>
+          <span className={styles.storeSummaryLabel}>Breakeven: </span>
+          <span>$1000 / month</span>
+        </div>
+      </div>
+
+      <div className={styles.socialContainer}>
+        {/* TEMPLATE!!! add appropriate links & social icons as needed*/}
```
jtmckibb, line 62: FYI @ArtyEmsee

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -9,16 +9,69 @@ interface Props {
...
+        <a href="#" className={classnames(styles.fa, "fa fa-twitter")} />
+        <a href="#" className={classnames(styles.fa, "fa fa-instagram")} />
+        <a href="#" className={classnames(styles.fa, "fa fa-facebook")} />
+        <a href="#" className={classnames(styles.fa, "fa fa-youtube")} />
+      </div>
+
+      <div className={styles.mapsContainer}>
+        {/* need to put in google API */}
```
jtmckibb, line 70: Is this free to use?

src/components/OwnerPanel/OwnerPanel.tsx
ath-chen: to import social media icons

src/components/OwnerPanel/OwnerPanel.tsx
ath-chen: I believe so, prob would use the basic maps embed API: 

https://developers.google.com/maps/documentation/embed/usage-and-billing

src/components/OwnerPanel/OwnerPanel.tsx
ArtyEmsee: Working on this now 😄

src/components/HeroBanner/styles.module.scss
ArtyEmsee: Isn't that a google material design standard? Either way we should probably add defined definitions that live in another file so that they can be consistent across the project

ArtyEmsee: Not suggesting it be part of this PR, I think it can come later

src/components/OwnerPanel/OwnerPanel.tsx
ArtyEmsee: we may want to consider having this live somewhere else so it could be imported into any component

ath-chen: moved to the index.html instead! all components can use now

PR #6, opened 2020-04-11
Athena/components

ath-chen:
- added stipe functionality --> added card section element for CC info
- changed payment & confirmation modal boxes to functional components (from class) 

to do: 
- update backend charges URL with the correct one when ready 
- fix css & formatting where needed


src/components/CheckoutForm/CheckoutForm.tsx
```diff
@@ -47,16 +51,17 @@ class CheckoutForm extends React.Component<Props, State> {
...
+  async getCheckoutSession() {
     const lineItems = this.state.lineItems;
     lineItems.description = `${this.state.formattedValue} ${this.props.option.toLowerCase()} for ${this.props.merchant}`;
     lineItems.amount = lineItems.amount * 100;
-    lineItems.quantity = 1; 
+    lineItems.quantity = 1;
 
     // should return stripe checkout session id once the endpoint is working
-    return axios.post('https://api.sendchinatownlove.com/charges', {
+    const res = await axios.post('http://localhost:3001/charges', {
```
jtmckibb, line 64: We should update this to be localhost when developing, but when we use yarn deploy, it replaces it with api.sendchinatownlove.com

src/components/ModalBilling/ModalBilling.tsx
```diff
@@ -1,9 +1,13 @@
 import * as React from 'react';
 import classnames from 'classnames';
-import styles from './styles.module.scss';
+import {Elements} from '@stripe/react-stripe-js';
+import {loadStripe} from '@stripe/stripe-js';
 
+import styles from './styles.module.scss';
 import ModalPayment from '../ModalPayment';
 
+const stripePromise = loadStripe('pk_test_5AByIibLOhR6WHL3Mwnmel3P00zm0pIDrD');
```
jtmckibb, line 9: This key should also be updated to use `pk_test_5AByIibLOhR6WHL3Mwnmel3P00zm0pIDrD` for development, but to use our real key when deploying

PR #8, opened 2020-04-12
Fix home line heights for mobile

ksin:
Looks a lil better than before

Screen Shot 2020-04-11 at 8 18 02 PM


jtmckibb:
If you have a sec can you drop in the new logo? nw if not

PR #10, opened 2020-04-13
Rob/addsApiCallsToComponents

ArtyEmsee:


src/utilities/api/interactionManager.ts
```diff
@@ -0,0 +1,81 @@
+import React from 'react';
+import axios from 'axios';
+import { CardElement, useStripe, useElements } from '@stripe/react-stripe-js';
+import { Merchant, PaymentParams } from './types';
+
+// const genericHeader = {
```
jtmckibb, line 6: don't think we need this

src/components/App/App.tsx
```diff
@@ -1,22 +1,45 @@
...
+      sellers,
+    });
+  }
+
+  render() {
+    return (
+      <Router history={history}>
+        <Switch>
+          <Route exact path="/" component={Home} />
+          <Route path="/story/:id" component={MerchantPage} />
```
jtmckibb, line 37: should this be :id/story ?

ArtyEmsee: I will add a TO DO to revisit this

PR #15, opened 2020-04-14
Athena/components

ath-chen:
- made responsive mobile pages 
- hid social buttons & photos tab & extra info section
- fixed stripe call with correct params


service-worker.js
```diff
@@ -12,7 +12,11 @@
  */
 
 importScripts(
+<<<<<<< HEAD
+  "/precache-manifest.e74ba72f14a210479a5ccdeac672c65b.js"
+=======
   'https://storage.googleapis.com/workbox-cdn/releases/4.3.1/workbox-sw.js'
+>>>>>>> 11d092887a48b469799cd76017634269a1937815
```
stanleyzhengnyc, line 19: this looks like a git conflict here

service-worker.js
```diff
@@ -12,7 +12,11 @@
  */
 
 importScripts(
+<<<<<<< HEAD
```
jtmckibb: @ath-chen looks like the merge conflict is still here

src/components/ModalPayment/ModalPayment.tsx
```diff
@@ -30,9 +30,9 @@ const ModalPayment = ({
   const payment: PaymentParams = {
     amount: Number(donatedAmt) * 100,
     currency: 'usd',
-    name: 'Gift Card',
+    item_type: purchaseType === 'donation' ? 'donation' : 'gift_card',
```
jtmckibb, line 33: Would be nice to abstract these into an enum on a class in the future. Don't need to do now

src/components/ModalConfirmation/styles.module.scss
```diff
@@ -21,3 +21,11 @@ form img {
...
   }
 }
+
+@media only screen and (max-width: 450px) {
+  .container {
+    width: 100%;
+    height: 100%;
+    overflow-x: hidden;
+  }
+}
```
jtmckibb, line 31: You should run `yarn format` before you push to format everything. You should also run eslint too

service-worker.js
ath-chen: updated

service-worker.js
ath-chen: updated

src/components/ModalPayment/ModalPayment.tsx
ath-chen: ya can update later on

PR #17, opened 2020-04-14
Rob/hides social buttons

ArtyEmsee:
Hides social buttons 👍


jtmckibb:
Looks like Athena already hid them :)

PR #21, opened 2020-04-14
Remove 'dark' theme

ksin:
Looks like kevin did something to make all the footers light

### DIFF
--- a/src/components/Home/Home.tsx
+++ b/src/components/Home/Home.tsx
@@ -18,7 +18,7 @@ const Home: React.SFC = () => {
           </Link>
         </div>
       </main>
-      <Footer theme={'dark'} />
+      <Footer />
     </div>
   );
 };


src/components/Home/Home.tsx
```diff
@@ -18,7 +18,7 @@ const Home: React.SFC = () => {
           </Link>
         </div>
       </main>
-      <Footer theme={'dark'} />
+      <Footer />
```
jtmckibb, line 21: FYI @kevinjtsai I noticed that you did this and it also caused the footer on the seller page to become stark white as opposed to the grey you have in designs. Was this intended?

kevinjtsai: Yep, I wanted to simplify the code and design by using a uniform header. I left the dark code logic for now, but we should probably clean it up. Happy to change it to a single color that works for landing page and product views.

kevinjtsai: can we remove the dark theme code?

function imgLogo(theme: Theme) {
  if (theme === 'dark') {
    return require(`./image/logo.svg`);
  }
  return require(`./image/logo.svg`);
}

PR #37, opened 2020-04-17
Johnt/progressbar

jt5301:
defaults to the red/purple color if there isn't a custom one

### DIFF
--- a/src/components/OwnerPanel/OwnerPanel.tsx
+++ b/src/components/OwnerPanel/OwnerPanel.tsx
@@ -44,7 +44,6 @@ class OwnerPanel extends React.Component<Props, State> {
   }
 
   render() {
-    console.log('in bar place', this.props)
     return (
       <section className={classnames(styles.container, this.props.className)}>
         <figure className={styles.ownerContainer}>
--- a/src/components/OwnerPanel/OwnerPanel.tsx
+++ b/src/components/OwnerPanel/OwnerPanel.tsx
@@ -13,6 +13,7 @@ interface Props {
   ownerName: string;
   sellerId: string;
   sellerName: string;
+  progressBarColor: string;
 }
 
 interface State {
@@ -33,7 +34,6 @@ class OwnerPanel extends React.Component<Props, State> {
     this.showModal = this.showModal.bind(this);
     this.hideModal = this.hideModal.bind(this);
   }
-
   showModal(event: any) {
     let button = event.target.value;
     this.setState({ show: true, purchaseType: button });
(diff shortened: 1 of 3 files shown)


src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -63,7 +63,8 @@ class OwnerPanel extends React.Component<Props, State> {
                 style={{
                   width: `${
                     (this.props.amountRaised / this.props.targetAmount) * 100
-                  }%`,
+                    }%`,
+                  backgroundColor: `${this.props.progressBarColor ? this.props.progressBarColor : ''}`
```
jtmckibb, line 67: What happens if the background color is `''`? Can we also do some validation that what is passed from the server is of form #XXXXXX` aka is a html hex color code?

jtmckibb: Mind adding a comment here that when the backgroundColor is `''` it defaults to what's defined in the CSS?

bvillaroman: I could do a regex pattern check if its a valid html hex and if it doesn’t pass, i can set it to our default color pattern.

jtmckibb: That would be perfect!

jtmckibb: It would be great if we had a util or or factory of source that all of the data from the server passed through so that all of the sanitization of server data could be put in one place. Plus it would be easy to unit test.

If you do that, it could be a follow-up. Just be sure to create a ticket on Trello so that we can track it.

jt5301: i see. since its nonblocking i simplified it and added a comment. as mentioned its not validated yet so if someone enters an invalid hex in defaults to gray. if its null it defaults to the color that was assigned in the css file

bvillaroman: @jt5301 should i add the regex on top of your code to validate the color or are you handling the whole progress bar

jt5301: @bvillaroman Feel free to add! I haven’t really taken ownership of the whole thing yet, was just doing color for now

jtmckibb: Let's merge this and add a separate ticket for sanitization

jtmckibb: Anyone feel free to pick it up and add it into the sprint: https://trello.com/c/l1bzIYjg/162-add-a-util-to-sanitize-server-responses

PR #39, opened 2020-04-18
Bruce/progressbar sanitizing

bvillaroman:
- added a sanitizer file that holds all the text sanitizers/validators. 
- first function validates an input if its a props hexcode 
- implemented hexcode validator OwnerPanel to apply to progress bar colors


src/utilities/sanitizers.ts
```diff
@@ -0,0 +1,5 @@
+export function validateHexcode(code: string){
+  const validPattern = new RegExp(/^#([A-Fa-f0-9]{6}|[A-Fa-f0-9]{3})$/)
```
jtmckibb, line 2: wow bless

PR #41, opened 2020-04-18
Remove js and css build static files

Yong-L:
These files should be generated at build time... right?

Correct me if I'm wrong 😄


jtmckibb:
I thought so! I was wondering why this was checked in

jtmckibb:
Gonna close this and redo it on the current dev

PR #43, opened 2020-04-20
Add Square payment flow

bvillaroman:


bvillaroman:
ok added the css fix and added the switch between stripe and square

PR #44, opened 2020-04-20
John/extrasellerinfo

jt5301:
Added seller extra info column below owner panel. Ends up like this:
Screen Shot 2020-04-19 at 10 59 35 PM

or with less info provided, like this:
Screen Shot 2020-04-19 at 10 59 26 PM

are those 5 the only extra pieces of info we will be adding to this section?


src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -44,6 +46,7 @@ class OwnerPanel extends React.Component<Props, State> {
   }
 
   render() {
+    console.log(this.props)
```
jtmckibb, line 49: nit should remove

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -97,6 +101,49 @@ class OwnerPanel extends React.Component<Props, State> {
...
+              borderTop: '1px solid #dedede',
+              borderBottom: '1px solid #dedede'
+            }}>
+            {Object.keys(this.props.extraInfo).map((current) => {
+              if (current === 'Website' || current == 'Menu') {
+                return (
+                  <>
+                    <p
+                      style={{
+                        fontWeight: 'bold'
```
jtmckibb, line 117: We should put this into a class

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -97,6 +101,49 @@ class OwnerPanel extends React.Component<Props, State> {
...
+                      <a href={`http://${this.props.extraInfo[current]}`}>{this.props.extraInfo[current]}</a>
+                    </p>
+                  </>
+                )
+              }
+              else {
+                return (
+                  <>
+                    <p style={{
+                      fontWeight: 'bold',
```
jtmckibb, line 129: Same here should be in a class

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -97,6 +101,49 @@ class OwnerPanel extends React.Component<Props, State> {
...
+                    }}>
+                      {`${current}: `}
+                      <span style={{
+                        fontWeight: 'normal'
```
jtmckibb, line 133: merp

src/components/OwnerPanel/styles.module.scss
```diff
@@ -88,3 +88,9 @@ $container-size: 160px;
   background-color: #dd678a;
   border: 1px solid #dd678a;
 }
+
+p {
```
jtmckibb, line 92: Instead of styling the actual tag, create a specific class for this so that you don't end up changing all of the nested p tags

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -13,6 +13,9 @@ interface Props {
   ownerName: string;
   sellerId: string;
   sellerName: string;
+  progressBarColor: string;
+  extraInfo: { [prop: string]: any }
```
jtmckibb, line 17: Where is this being defined/where is it coming from? iirc @juliexxia's PR hasn't merged yet

jt5301: yeah, I'm grabbing it from SellerPage and passing it in as extraInfo = {seller.extra_seller_info} as a standin. I based it off how location model got loaded with the seller object and was assuming it'll be the same with the extra info model. My bad if I should've held off, I can change once the model is merged.

jtmckibb: Ah got it. I think the extra info will be directly on the seller model: https://github.com/sendchinatownlove/ruby/pull/29

Not like location which is a separate object. But no this is totally good!

jtmckibb: As long as the model in the backend stays on the seller model, you can model it however you want in the frontend. I think encapsulating it like that is fine :)

jt5301: Got it!

juliexxia: sendchinatownlove/ruby#29 is submitted! extra info is directly on the seller model like Justin said, not in its own model.

jt5301: Ty Julie! Will make changes on my end

jt5301: Hi @juliexxia, I'm getting an error pulling+ migrating after the latest merge saying there's an undefined model called extra_seller_infos. I haven't picked up enough/any rails to fix the issue on my end, do you have any suggestions on what I can do?
Screen Shot 2020-04-21 at 12 46 46 AM

Not sure if I was the only one getting it, so didn't post on slack

jtmckibb: you have to migrate your database because she added new fields to the Sellers table.
```
rails db:create # only if you haven't done this before
rails db:migrate # to run a migration
```

jt5301: I got that error after dropping ->recreating->migrate. It must be on my end only, I'll do some digging
Screen Shot 2020-04-21 at 12 56 06 PM

jtmckibb: Sent you a message on Slack

jt5301:
ready to merge, checking to see if you wanted to look over a final time

src/components/SellerPage/SellerPage.tsx
```diff
@@ -52,6 +50,13 @@ const SellerPage: React.FC<{}> = () => {
               imageSrc={sampleMerchant.ownerInfo.imageSrc}
               sellerName={seller.name}
               progressBarColor={seller.progress_bar_color}
+              extraInfo={{
```
jtmckibb, line 53: Nice have you tested this with the backend?

src/components/OwnerPanel/styles.module.scss
```diff
@@ -88,3 +88,19 @@ $container-size: 160px;
...
+  line-height: 1.75em;
+  letter-spacing: 0.015em;
+}
+
+.extraInfoKey{
+  font-weight: bold
+}
+
+.extraInfoValue{
+  font-weight:normal
```
jtmckibb, line 103: nit needs a space after this. Before you merge, can you run `yarn format` and `yarn eslint --fix .`

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -87,13 +88,50 @@ const OwnerPanel = (props: Props) => {
           </button>
         )}
       </div>
+      {Object.keys(props.extraInfo).length !== 0 ?
```
jtmckibb, line 91: I think if/else would be easier to read here

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -87,13 +88,50 @@ const OwnerPanel = (props: Props) => {
...
+        <div
+          style={{
```
jtmckibb, line 93: This needs to be in a class. We shouldn't have styles outside of our css

src/components/SellerPage/SellerPage.tsx
jt5301: yep, works with extra info columns

src/components/OwnerPanel/styles.module.scss
jt5301: done

src/components/OwnerPanel/OwnerPanel.tsx
jt5301: sorry, can you clarify? i thought jsx doesn't like if/else?

jtmckibb:
Oh it might not 🙃

I didn’t know — I’m not a React eng

PR #45, opened 2020-04-21
Bruce/restructure payment modal

bvillaroman:


src/components/ModalPayment/SquareModal.tsx
```diff
@@ -61,117 +56,113 @@ const ModalPayment = ({
...
+          applicationId={applicationId}
+          locationId={locationId}
+          cardNonceResponseReceived={cardNonceResponseReceived}
+          formId="SPF"
+          apiWrapper=""
+        >
+          <SimpleCard/>
+          <br/>
+          <h3>Checkout details</h3>
+          {/* TODO(jmckibben): This should be replaced with Seller.name */}
```
jtmckibb, line 125: Thanks for doing this! You can remove this TODO

src/components/ModalPayment/SubmissionButton.tsx
```diff
@@ -2,12 +2,13 @@ import React, { useContext } from 'react';
...
 type Props = {
-  isChecked: boolean;
+  canSubmit: boolean;
 };
 
-const SubmissionButton = ({isChecked}: Props) => {
+const SubmissionButton = ({canSubmit}: Props) => {
   const context = useContext(Context);
  
+  console.log(context)
```
jtmckibb, line 11: Forgot to delete this ^^

src/components/ModalPayment/SubmissionButton.tsx
```diff
@@ -2,12 +2,13 @@ import React, { useContext } from 'react';
 import { Context } from 'react-square-payment-form';
 
 type Props = {
-  isChecked: boolean;
+  canSubmit: boolean;
```
jtmckibb, line 5: nice!

src/utilities/hooks/ModalPaymentContext/types.ts
```diff
@@ -0,0 +1,22 @@
...
+  amount: number;
+  customInput: boolean;
+  modalView: number;
+}
+
+export const defaultState: ModalPaymentState =  {
+  amount: 0,
+  customInput: false,
+  modalView: -1
+}
```
jtmckibb, line 22: nit new line

src/components/ModalPayment/SquareModal.tsx
```diff
@@ -61,117 +56,113 @@ const ModalPayment = ({
...
+          <span> {purchaseTypePhrase} of <b>${amount}</b> to ${sellerName} </span>
```
jtmckibb, line 126: shouldn't have $ before sellerName

src/components/ModalAmount/styles.module.scss
```diff
@@ -0,0 +1,61 @@
...
+
+.errorMessage {
+  color: red;
+  font-size: 13px;
+}
+
+@media only screen and (max-width: 799px) {
+  .donationsContainer {
+    width: 85%;
+    height: 75vh;
```
jtmckibb, line 42: not super sure what vh is

src/components/Modal/styles.module.scss
```diff
@@ -1,5 +1,5 @@
 .donationsContainer {
-  height: 500px;
+  height: auto;
```
jtmckibb, line 2: do we need this here?

src/utilities/hooks/ModalPaymentContext/reducers.ts
```diff
@@ -0,0 +1,31 @@
...
+  CLEAR_FORMS
+} from "./constants"
+import { defaultState, ModalPaymentState } from "./types"
+
+export interface Action {
+  type: string,
+  payload: any
+}
+
+const ModalPaymentReducer = (state: ModalPaymentState, action: Action) => {
```
jtmckibb, line 15: This is so much better thank you for doing this!!

src/utilities/hooks/ModalPaymentContext/types.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ */
+
+export type ModalPaymentState = {
+  amount: number;
+  customInput: boolean;
+  modalView: number;
+}
+
+export const defaultState: ModalPaymentState =  {
```
jtmckibb, line 18: I love it

src/components/ModalPayment/SquareModal.tsx
```diff
@@ -61,117 +56,113 @@ const ModalPayment = ({
...
+          const newErrors = errorMessages.length > 0 ? [...errorMessages,responseErrors.map((error: { detail: string }) => error.detail) ] : responseErrors.map((error: { detail: string }) => error.detail)
+          setErrorsMessages(newErrors)
         }
       })
   }
 
   const applicationId = process.env.REACT_APP_SQUARE_APPLICATION_ID ? process.env.REACT_APP_SQUARE_APPLICATION_ID : ""
   const locationId = process.env.REACT_APP_SQUARE_SANDBOX_LOCATION_ID ? process.env.REACT_APP_SQUARE_SANDBOX_LOCATION_ID : ""
 
+  const canSubmit = isChecked && name.length > 0 && email.length > 0;
```
jtmckibb, line 77: also if there are no Square errors

src/components/Modal/Modal.tsx
```diff
@@ -1,161 +1,42 @@
...
+  return (
+    <div
+      id="donation-form"
+      className={classnames(styles.donationsContainer, 'modalForm--form')}
+      style={{ display: modalView > -1 ? 'block' : 'none' }}
+    >
+      <button className={'closeButton--close'} onClick={closeModal} > 
+        ×
+      </button>
+      {modalView === 0 && <ModalAmount {...props}/>}
```
jtmckibb, line 35: Does JavaScript have enums?

src/components/ModalConfirmation/styles.module.scss
```diff
@@ -2,7 +2,7 @@
   height: 650px;
 }
 
-form img {
+img {
```
jtmckibb, line 5: Would prefer we use classes, but this is fine

src/components/ModalAmount/styles.module.scss
```diff
@@ -0,0 +1,61 @@
...
+
+.customAmt {
+  width: 250px;
+}
+
+.nextBtn {
+  position: relative;
+  float: right;
+  right: 0px;
+  bottom: -25px;
```
jtmckibb, line 27: nit: bad practice to use negative heights, but we can amend later

src/components/ModalPayment/SquareModal.tsx
```diff
@@ -61,117 +56,113 @@ const ModalPayment = ({
...
       .then((res) => {
         if (res.status === 200) {
-          setIsShown(true)
-        } else {
-          setIsShown(false)
+          dispatch({type: SET_MODAL_VIEW, payload: 2})
+        } 
+      })
+      .catch((err) => {
+        if( err.response
```
jtmckibb, line 63: this is pretty gnarly — do you mind running yarn format && yarn eslint --fix before you merge?

src/components/ModalPayment/SquareModal.tsx
```diff
@@ -61,117 +56,113 @@ const ModalPayment = ({
...
+          && err.response.data
+          && err.response.data.errors
+          && err.response.data.errors.length > 0){          
+          const responseErrors = err.response.data.errors
+          const newErrors = errorMessages.length > 0 ? [...errorMessages,responseErrors.map((error: { detail: string }) => error.detail) ] : responseErrors.map((error: { detail: string }) => error.detail)
```
jtmckibb, line 68: In the future we probably want to set custom error messages either on the client or server vs using Square's. We can keep this in mind as a follow up

src/components/Modal/Modal.tsx
ArtyEmsee: TypeScript does

src/components/Modal/Modal.tsx
```diff
@@ -1,161 +1,42 @@
...
+import ModalAmount from '../ModalAmount';
+import {SquareModal} from '../ModalPayment';
+import ModalConfirmation from '../ModalConfirmation';
+import { useModalPaymentState, useModalPaymentDispatch } from "../../utilities/hooks/ModalPaymentContext/context"
+import { CLOSE_MODAL } from "../../utilities/hooks/ModalPaymentContext/constants"
 
 interface Props {
   purchaseType: string;
-  handleClose: (event: React.MouseEvent<HTMLButtonElement>) => void;
-  show: boolean;
```
ArtyEmsee, line 12: I'd rename this to `shouldShow` or something, show implies action to me

src/components/Modal/Modal.tsx
```diff
@@ -1,161 +1,42 @@
...
   sellerId: string;
   sellerName: string;
 }
 
-interface State {
```
ArtyEmsee, line 17: nit: doesn't really need to be an interface since state isn't a class.

src/components/Modal/Modal.tsx
```diff
@@ -1,161 +1,42 @@
...
-  amount: number;
-  customInput: boolean;
-  close: boolean;
-  next: boolean;
-  showBillModal: boolean;
-}
-
-const ModalPaymentBox: any = process.env.REACT_APP_USE_STRIPE === "TRUE" ? StripeModal : SquareModal ;
-
-const stripePK = process.env.REACT_APP_STRIPE_PUBLISHABLE_KEY!;
```
ArtyEmsee, line 27: This null assertion at compile time here is a little problematic since it can lead to run time errors if it fails to get the key for some reason. We should probably add a to do or something to come back and make this more fail safe.

src/components/Modal/Modal.tsx
```diff
@@ -1,161 +1,42 @@
...
-      <form
-        id="donation-form"
-        className={classnames(styles.donationsContainer, 'modalForm--form')}
-        style={{ display: this.props.show ? 'block' : 'none' }}
-      >
-        <button
-          className={'closeButton--close'}
-          onClick={this.props.handleClose}
-        >
-          {' '}
```
ArtyEmsee, line 71: nit: imo ` ${x} ` is cleaner than this pattern for spaces.

src/components/Modal/Modal.tsx
ArtyEmsee: I agree this should live as an enum somewhere

src/components/ModalAmount/styles.module.scss
ArtyEmsee: I believe it's viewport height? We should probably either avoid using it interchangeably w percentage or only use it.

src/components/Modal/styles.module.scss
ArtyEmsee: imo we can utilize flexbox to standardize container behavior

src/components/ModalConfirmation/ModalConfirmation.tsx
```diff
@@ -3,39 +3,26 @@ import classnames from 'classnames';
...
+    <div className={styles.paymentContainer}>
       <h2>Thank you for your donation!</h2>
       <p>We appreciate your support for small businesses</p>
 
       <img className={styles.image} src={confirmationPic} alt="Logo" />
 
       <button
         className={classnames(styles.finishBtn, 'modalButton--filled')}
-        onClick={handleClose}
+        onClick={() => dispatch({type: CLOSE_MODAL, payload: undefined})}
```
ArtyEmsee, line 21: shouldn't need to send a payload of undefined?

src/components/ModalPayment/SquareModal.tsx
```diff
@@ -6,47 +6,42 @@ import 'react-square-payment-form/lib/default.css'
 import styles from './styles.module.scss';
 import SubmissionButton from './SubmissionButton';
 import {makeSquarePayment, SquarePaymentParams, Buyer} from "../../utilities/api"
-import ModalConfirmation from '../ModalConfirmation';
+
+import { useModalPaymentState, useModalPaymentDispatch } from "../../utilities/hooks/ModalPaymentContext/context"
+import { SET_MODAL_VIEW } from "../../utilities/hooks/ModalPaymentContext/constants"
```
ArtyEmsee, line 11: Nice, glad people are utilizing this

ArtyEmsee, line 11: Though we should probably look at a way to avoid doing all this stuff with relative paths, there are packages available

src/components/ModalPayment/SquareModal.tsx
```diff
@@ -61,117 +56,113 @@ const ModalPayment = ({
...
-        } else {
-          setIsShown(false)
+          dispatch({type: SET_MODAL_VIEW, payload: 2})
+        } 
+      })
+      .catch((err) => {
+        if( err.response
+          && err.response.data
+          && err.response.data.errors
+          && err.response.data.errors.length > 0){
```
ArtyEmsee, line 66: I don't think you need to go down the whole object chain to test for this

src/components/ModalPayment/SquareModal.tsx
ArtyEmsee: yeah we'll likely want to expand this to test email validity, there are some regex ways to do this or we can look at other options

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -23,48 +25,35 @@ interface State {
...
-  constructor(props: Props) {
-    super(props);
-    this.state = {
-      // these will be props fix throughout when passed thru
-      show: false,
-      purchaseType: '',
-    };
-    this.showModal = this.showModal.bind(this);
-    this.hideModal = this.hideModal.bind(this);
+const OwnerPanel =(props: Props) => {
```
ArtyEmsee, line 28: this looks like it needs to be prettified

src/components/ModalConfirmation/ModalConfirmation.tsx
bvillaroman: i made it so that all actions require a payload, undefined dispatches have a specific action in changing the the state that don't require a payload

PR #58, opened 2020-04-23
future/url-resolution

stanleyzhengnyc:


public/CNAME
```diff
@@ -1 +1 @@
-www.sendchinatownlove.com
\ No newline at end of file
+merchant.sendchinatownlove.com
```
jtmckibb, line 1: Should this be www.merchant?

PR #59, opened 2020-04-23
extrainfo fix, t&c hit box is bigger

jt5301:
extra info section hides null. Also picked up the t&c hitbox minor fix, hovering underlines the text and you can click anywhere in the container. open to suggestions on anything better!
Screen Shot 2020-04-23 at 1 01 20 PM

### DIFF
--- a/src/components/ModalPayment/styles.module.scss
+++ b/src/components/ModalPayment/styles.module.scss
@@ -14,7 +14,7 @@
 
 .row,
 .inputRow,
-.tc {
+.termsAndConditions {
   display: flex;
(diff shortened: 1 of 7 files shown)


src/components/ModalPayment/styles.module.scss
```diff
@@ -13,14 +13,19 @@
 }
 
 .row,
-.inputRow {
+.inputRow,
+.tc {
```
jtmckibb, line 17: Could you write out termsAndConditions? Makes it easier to understand

PR #67, opened 2020-04-24
Update thank you page copy

jtmckibb:


src/components/ModalConfirmation/ModalConfirmation.tsx
```diff
@@ -6,14 +6,23 @@ import confirmationPic from './chinatown-logo.png';
 import { useModalPaymentDispatch } from '../../utilities/hooks/ModalPaymentContext/context';
 import { CLOSE_MODAL } from '../../utilities/hooks/ModalPaymentContext/constants';
 
-const ModalConfirmation: React.SFC = () => {
+type Props = {
+  purchaseType: string;
+  sellerId: string;
+  sellerName: string;
```
ath-chen, line 12: you could just pass in sellerName since the other props aren't being used but looks good! we can fix that later

PR #72, opened 2020-04-25
Athena/components

ath-chen:
- added idempotentKey

### DIFF
--- a/src/components/Modal/Modal.tsx
+++ b/src/components/Modal/Modal.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import classnames from 'classnames';
+import { v4 as uuid } from 'uuid';
 import styles from './styles.module.scss';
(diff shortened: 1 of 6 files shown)


src/components/Modal/Modal.tsx
```diff
@@ -16,6 +17,8 @@ interface Props {
   sellerName: string;
 }
 
+const idempotentKey = uuid();
```
jtmckibb, line 20: nit, but we should call this `idempotencyKey`

src/utilities/api/types.ts
```diff
@@ -12,6 +12,7 @@ export type SquarePaymentParams = {
   item_type: string;
   quantity: number;
   seller_id: string;
+  idempotentKey: string;
```
jtmckibb, line 15: `idempotencyKey`

PR #73, opened 2020-04-25
changed var name

ath-chen:


src/utilities/api/types.ts
```diff
@@ -12,7 +12,7 @@ export type SquarePaymentParams = {
   item_type: string;
   quantity: number;
   seller_id: string;
-  idempotencyKey: string;
+  idempotency_key: string;
```
jtmckibb, line 15: this should be on the Buyer and not the line_item

ath-chen: OH MY BAD, moving it now

PR #74, opened 2020-04-25
Move seller_id for square_payment to charges body

Yong-L:
Reference


Yong-L:
Reference: https://github.com/sendchinatownlove/ruby/pull/68

src/utilities/api/interactionManager.ts
```diff
@@ -87,6 +88,7 @@ export const makeSquarePayment = async (
         line_items: [payment],
         email,
         name,
+		seller_id: sellerId
```
jtmckibb, line 91: you should run yarn format and yarn eslint --fix before you merge

PR #77, opened 2020-04-26
Athena/components

ath-chen:
- added modal billing
- can the backend take in address info yet? currently sending it in buyer obj 

Screen Shot 2020-04-26 at 4 56 47 PM


jtmckibb:
I (think) that we should be using “createVerificationDetails” here: https://square.github.io/react-square-payment-form/docs/paymentform

https://developer.squareup.com/docs/sca-overview?q=sca

and if I am understanding this correctly, you’d pass the nonce + the verificationToken to the backend, and we’d use that to charge the card

src/components/MerchantsPage/MerchantsPage.tsx
```diff
@@ -32,8 +32,8 @@ const MerchantsPage: React.FC<{}> = () => {
         <NavBar />
 
         <div className={styles.merchantsContainer}>
-          {sellers.map((store: any) => (
-            <MerchantCard storeInfo={store} />
+          {sellers.map((store: any, idx: any) => (
```
bvillaroman, line 35: using the index isn't the best for assigning keys for child components because react will confuse children elements whenever it changes dynamically and might create some sorting issues down the lane. More in depth explanation: https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

jtmckibb: Good callout

src/components/ModalConfirmation/styles.module.scss
```diff
@@ -1,5 +1,6 @@
 .container {
-  max-height: 500px;
+  height: 75vh;
```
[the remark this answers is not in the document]
ath-chen: this is vh bc its taking the viewport height regardless of parent size. % is relative to the parent container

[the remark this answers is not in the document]
jtmckibb: I won't argue with the css guru

src/components/MerchantsPage/MerchantsPage.tsx
ath-chen: good call! fixed throughout :)

jtmckibb:
@ath-chen I hate to do this but I think I didn't do enough reading about SCA for Square. It requires  buyers to enter their phone number and receive a text message where they'll authorize a payment. I think it might be better we don't do this for now — and if fraud becomes a bigger fear we can implement this. I'll look into why Square is allowing for us to process CCs with the wrong zip though. I'm going to decide this PR in the mean time. We can put it on ice for now. Sorry again!!

PR #79, opened 2020-04-28
Disable button to prevent duplicate donations/purchages

danthaman44:
Demo: https://drive.google.com/open?id=1m51ZlR56oLG4DVIsVw8YMa9r6LXxVKrJ

Disables the button right after hitting submit. The button can be re-enabled by clicking on any input field.

### DIFF
--- a/src/components/ModalPayment/SubmissionButton.tsx
+++ b/src/components/ModalPayment/SubmissionButton.tsx
@@ -1,24 +1,31 @@
-import React, { useContext } from 'react';
+import React, { useContext, useState } from 'react';
 import { Context } from 'react-square-payment-form';
 
 type Props = {
   canSubmit: boolean;
+  setChecked: (isChecked: boolean) => void;
 };
 
-const SubmissionButton = ({ canSubmit }: Props) => {
+const SubmissionButton = ({ canSubmit, setChecked }: Props) => {
   const context = useContext(Context);
+  var [submittable] = useState(false);
 
   const handleSubmit = (evt: { preventDefault: () => void }) => {
     evt.preventDefault();
     context.onCreateNonce();
   };
 
+  submittable = canSubmit;
+
   return (
     <button
       type="button"
       className={'modalButton--filled'}
-      onClick={handleSubmit}
-      disabled={canSubmit === false}
+      onClick={(e) => {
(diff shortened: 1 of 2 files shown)


src/components/ModalPayment/SubmissionButton.tsx
```diff
@@ -7,18 +7,22 @@ type Props = {
...
 
+  submittable = canSubmit;
+
   return (
     <button
       type="button"
       className={'modalButton--filled'}
       onClick={handleSubmit}
-      disabled={canSubmit === false}
+      disabled={!submittable}
```
jtmckibb, line 25: Does this also work for Enter to submit? Do we even support that?

src/components/ModalPayment/SubmissionButton.tsx
```diff
@@ -7,18 +7,22 @@ type Props = {
 
 const SubmissionButton = ({ canSubmit }: Props) => {
   const context = useContext(Context);
+  var [submittable, setSubmittable] = useState(false);
 
   const handleSubmit = (evt: { preventDefault: () => void }) => {
     evt.preventDefault();
     context.onCreateNonce();
+    setSubmittable(false);
```
jtmckibb, line 15: When do we set this to true?

src/components/ModalPayment/SubmissionButton.tsx
danthaman44: It doesn't look like we support that atm. Thats a good idea for a feature though, I can try adding it in another PR

src/components/ModalPayment/SubmissionButton.tsx
danthaman44: Clicking on any of the inputs in the modal will re-render button and the button will be enabled again (assuming the criteria around name, email, and terms/conditions are met)

danthaman44: Scratch that, it doesn't work. Gonna rework this PR

danthaman44: Got it to work with some help from @ath-chen ! Check out the updated video https://drive.google.com/open?id=1m51ZlR56oLG4DVIsVw8YMa9r6LXxVKrJ

PR #85, opened 2020-05-02
Johnt/mobile bug

jt5301:
gave header some padding on top + flex box. Also got rid of width limit that was messing with T&C section.
Screen Shot 2020-05-01 at 9 25 33 PM

### DIFF
--- a/src/components/ModalPayment/SquareModal.tsx
+++ b/src/components/ModalPayment/SquareModal.tsx
@@ -83,14 +83,14 @@ const ModalPayment = ({
           const newErrors =
             responseErrors.length > 0
               ? responseErrors.map(
-                (error: { code: string; detail: string }) => {
-                  if (hasKey(SquareErrors, error.code)) {
-                    return SquareErrors[error.code];
-                  } else {
-                    return error.detail;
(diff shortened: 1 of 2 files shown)


src/components/ModalPayment/styles.module.scss
```diff
@@ -14,12 +14,22 @@
   }
 }
 
+@media screen and (max-width: 450px) {
+  .headerView {
+    padding-top: 25px;
+    display: flex;
```
jtmckibb, line 20: Why not make it always use flexbox with text-align center?

jt5301: good point :sweat_smile: i'll bring it out

src/components/ModalPayment/SquareModal.tsx
```diff
@@ -193,14 +195,14 @@ const ModalPayment = ({
               exchanged for this donation.
             </p>
           ) : (
-            <p>
-              By proceeding with your purchase, you understand that the gift
+              <p>
```
jtmckibb, line 198: Be sure to run `yarn eslint --fix && yarn format` before merging!

PR #86, opened 2020-05-02
Athena/merchants page

ath-chen:
- updated merchants directory to latest design 
- integrated with backend API calls for the contribution bar 

todo: 
- include "last donation made" when available

demo: 
https://drive.google.com/open?id=19phecB_1M18J43QU252-6OiW7uN8rO91


src/components/MerchantsPage/ContributionBar.tsx
```diff
@@ -0,0 +1,60 @@
+import * as React from 'react';
+import styled from 'styled-components';
+
+interface Props {
+  totalDonations: number;
+  totalVouchers: number;
```
jtmckibb, line 6: Even though they're called vouchers externally, can we call them gift cards internally because that's how they're modeled and I don't wanna change the name of our datatbase tables

src/components/MerchantsPage/DescriptionBox.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from 'react';
+import styled from 'styled-components';
+import styles from './styles.module.scss';
+import merchantIcon from './images/merchantIcon.png';
+
+const Container = styled.div`
```
jtmckibb, line 6: Interesting I have never seen this notation before. Is there a reason why we're doing this as opposed to putting it into an scss?

src/components/MerchantsPage/styles.module.scss
```diff
@@ -33,18 +35,87 @@
...
+  }
+}
+
+.storeInfo {
+  grid-row: 2;
+  margin: 30px 0px;
+  // border: 1px solid black;
+
+  @include for-tablet-landscape-up {
+    // display: flex;
```
jtmckibb, line 111: merp

src/components/MerchantsPage/MerchantCard.tsx
```diff
@@ -1,9 +1,23 @@
...
+  color: #46accc;
+`;
+
+const Summary = styled.p`
+  display: -webkit-box;
+  height: 45px;
+  -webkit-line-clamp: 2;
+  -webkit-box-orient: vertical;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
jtmckibb, line 19: Could we not use a scss macro for overflow: hidden and text-overflow: ellipsis?

src/components/MerchantsPage/ContributionBar.tsx
ath-chen: yea! i can change this back

src/components/MerchantsPage/DescriptionBox.tsx
ath-chen: its cleaner and easier to know which css pieces are associated to which! im trying to break out our components into smaller pieces now and i don't to have a css file for each. 

also if there's a css code that is only applicable to that particular component, doesn't make sense to put it all in one shared css file. But I still do have css files in case there's code that can be used for multiple components

src/components/MerchantsPage/MerchantCard.tsx
ath-chen: if we do that ^ it's only for one line of text which looked kinda weird. 

https://css-tricks.com/almanac/properties/l/line-clamp/

src/components/MerchantsPage/DescriptionBox.tsx
jtmckibb: Wow how interesting. It breaks a lot of very long-standing web idioms of keeping CSS styles separate from function and layout, but I trust your judgment!

bvillaroman: While I do love styled-components and use it personally and for work, I do believe in keeping our project consistent over cleanliness.  If we are going to move on with using styled-components, then we're going to have to do a slow but big refactor in the current code in styling, otherwise having only a couple of components using this library when it could be done in css. 

That being said, i do prefer the styled-components way of styling and would love to do this refactor with athena's help after we finish responsiveness for the card forms.

src/components/MerchantsPage/DescriptionBox.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const UL = styled.ul`
+  padding: 0 5px 0 20px;
+`;
+
+const BP = styled.li`
+  text-align: left;
+  padding: 5px 0;
+`;
+
```
bvillaroman, line 22: perhaps better name conventions for the tags for readability? I also prefer having the styles below the components just because if i ever want to look at a component, more often than not i'm going to be looking at it to change the react code itself instead of its styles.

src/components/MerchantsPage/MerchantsPage.tsx
```diff
@@ -1,45 +1,80 @@
...
 const MerchantsPage: React.FC<{}> = () => {
   const [sellers, setSellers] = useState<any | null>();
+  const [filter, setFilter] = useState<any | null>();
 
   const fetchData = async () => {
     const result = await getSellers();
-    setSellers(result.data);
+    await setSellers(result.data);
+    await setFilter(result.data);
   };
```
bvillaroman, line 21: Is there a reason why theres an await for the useState hooks?

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -147,29 +126,6 @@ const OwnerPanel = (props: Props) => {
...
-          <span>3 months</span>
-        </div>{' '}
-        <br />
-        <div>
-          <span className={styles.storeSummaryLabel}>Breakeven: </span>
-          <span>$1000 / month</span>
-        </div>
-      </div> */}
-
       {/* hide social links until needed */}
```
bvillaroman, line 129: great cleanup

src/components/MerchantsPage/DescriptionBox.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <List>
+      <Bullets>Asian immigrant-owned small businesses</Bullets>
+      <Bullets>
+        Not fluent in English, run cash-only establishments, and not tech-savvy
+      </Bullets>
+      <Bullets>
+        Currently struggling to pay rent, utility bill and employee wages
+      </Bullets>
+      <Bullets>
+        Do not have websites, online delivery service, or gift cards available
```
jtmckibb, line 19: Should be vouchers

src/components/MerchantsPage/MerchantNavBar.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const setStoreType = (type: any, e: any) => {
+    e.preventDefault();
+    setSelected(type);
+    filterStoreType(type);
+  };
+
+  const storeTypes = [
+    { value: 'all', text: 'All' },
+    { value: 'Bakery', text: 'Bakery' },
+    { value: 'Chinese', text: 'Restaurant' },
```
jtmckibb, line 21: I'll update these in the DB once we get our first non-Chinese restaurant

src/components/MerchantsPage/MerchantsPage.tsx
```diff
@@ -1,43 +1,97 @@
...
+              <h2 style={{ fontWeight: 'bolder' }}>Our Chinatown</h2>
+              <br />
+              <p>
+                We are providing an online platform to low-tech, cash-only,
+                Asian-owned small businesses that have been disproportionately
+                impacted by COVID-19.
+              </p>
+              <p>
+                Support local merchants by making a donation or purchasing a
+                gift card from them.
```
jtmckibb, line 69: voucher

PR #89, opened 2020-05-03
added mobile nav bar

bvillaroman:


src/components/Navbar/NavBar.tsx
```diff
@@ -1,25 +1,82 @@
...
+  };
+
+  return (
+    <div className={styles.navBarContainer}>
+      <img className={styles.image} src={sclLogo} alt="Logo" />
+      {hamburgerOpen ? (
+        showCompactMenu()
+      ) : (
+        <div className={styles.navLinksContainer}>
+          <Link to="/" className={styles.navLink}>
```
jtmckibb, line 69: This should link to https://sendchinatownlove.com/

src/components/Navbar/NavBar.tsx
```diff
@@ -1,25 +1,82 @@
...
+            HOME
+          </Link>
+          <Link to="/sellers" className={styles.navLink}>
```
jtmckibb, line 72: We should update this route to be merchants/

src/components/Navbar/NavBar.tsx
```diff
@@ -1,25 +1,82 @@
...
+            MERCHANTS
+          </Link>
+          <Link to="/about" className={styles.navLink}>
```
jtmckibb, line 75: I don't think we have this page implemented in Weebly so you can remove it here.

jtmckibb:
Looks good with some nits

jtmckibb:
I also realized that our 404 page has the wrong links in it XD

src/components/Navbar/NavBar.tsx
bvillaroman: the / is wherever the index lives in (when deployed should be sendchinatownlove.github.io)

src/components/Navbar/NavBar.tsx
bvillaroman: the path for all of our merchants is currently /sellers right now. I could change the path, but i would rather do that in a new branch

src/components/Navbar/NavBar.tsx
jtmckibb: yeah but our home page is hosted by Squarespace which lives under the main domain, while this lives under the merchants subdomain (which is hosted by GitHub). We plan to delete the existing home page under merchants. It shouldn't be accessible and should be replaced with a 404

bvillaroman: ah ok, will change here then

PR #91, opened 2020-05-03
Updated copy from gift card to voucher

danthaman44:
Screen Shot 2020-05-02 at 11 53 34 PM
Screen Shot 2020-05-02 at 11 55 07 PM


danthaman44:
Gonna leave use of `gift card` variables unchanged for now, we can migrate that over later

jtmckibb:
Definitely leave the usage of gift card variables and classnames, because that's what we call them on the server and in our datables. It'll be easier to keep it that way. However copy/externally we can update things to voucher

PR #93, opened 2020-05-04
Added # supporters and tooltip showing donation/giftcard breakdown

Jeremy-Xue:
Owner panel now supports # donations/gift cards sold as well as # supporters. Should display on hover on web and on tap on mobile web over # supporters.

### DIFF
--- a/src/components/OwnerPanel/OwnerPanel.tsx
+++ b/src/components/OwnerPanel/OwnerPanel.tsx
@@ -5,12 +5,20 @@ import Modal from '../Modal';
 import { useModalPaymentDispatch } from '../../utilities/hooks/ModalPaymentContext/context';
 import { SET_MODAL_VIEW } from '../../utilities/hooks/ModalPaymentContext/constants';
 import defaultOwnerImage from './assets/female.svg';
+import { withStyles, Theme } from '@material-ui/core/styles';
+import Tooltip from '@material-ui/core/Tooltip';
+
 
 interface Props {
   imageSrc: string;
   className?: string;
   amountRaised: number;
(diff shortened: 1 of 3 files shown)


Jeremy-Xue:
Should the user-facing language for gift cards be changed to vouchers?

bvillaroman:
> Should the user-facing language for gift cards be changed to vouchers?

it should already be in dev unless i missed some text about it. Do a rebase and see if it changes gift cards to vouchers

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -41,6 +49,16 @@ const OwnerPanel = (props: Props) => {
     return 100;
   };
```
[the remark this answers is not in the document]
ath-chen: Hmm I would prob just pass in seller since it's only going down one level. But if we end up creating new components and need to pass the props down more layers, then we should def use context

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -138,8 +182,8 @@ const OwnerPanel = (props: Props) => {
           })}
         </div>
       ) : (
-        ''
-      )}
+          ''
+        )}
```
jtmckibb, line 186: is this right? did you run `yarn format` and `yarn eslint --fix .`

Jeremy-Xue: just ran it, thanks for the catch!

jtmckibb:
Is this good to be merged?

PR #98, opened 2020-05-07
Updated seller modal hover states

danthaman44:
Talked to Nanxi offline, the red is only supposed to be a hover state for the buttons. 
Reverted the base color to black/white and implemented the hover state.
modal-hover


jtmckibb:
I really appreciate the gifs/screenshots

PR #99, opened 2020-05-07
Added social share icons to seller page

danthaman44:
Social icons implemented via https://github.com/nygardk/react-share
Screen Shot 2020-05-07 at 12 53 52 AM
Screen Shot 2020-05-07 at 9 35 27 PM

### DIFF
--- a/src/components/OwnerPanel/OwnerPanel.tsx
+++ b/src/components/OwnerPanel/OwnerPanel.tsx
@@ -6,6 +6,12 @@ import Modal from '../Modal';
 import ProgressBar from '../ProgressBar';
 import styles from './styles.module.scss';
 import defaultOwnerImage from './assets/female.svg';
+import { SocialIcon } from 'react-social-icons';
+import {
+  EmailShareButton,
+  FacebookShareButton,
+  TwitterShareButton,
+} from 'react-share';
 
 interface Props {
   imageSrc: string;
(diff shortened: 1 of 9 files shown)


src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -45,6 +53,9 @@ const OwnerPanel = (props: Props) => {
     return props.extraInfo[current] != null;
   });
 
+  const location = window.location.href;
+  const facebookQuote = 'Help raise money for ' + props.sellerName;
```
danthaman44, line 57: I just improvised this on the spot...any suggestions?

nanxiy: gonna slack you the icons we've been using

jtmckibb: Chinatown has been suffering from the effects of COVID. Help keep Chinatown alive by donating/buying a gift card! Check out <Insert Merchant Name>'s page:

idk something like that?

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -170,13 +181,31 @@ const OwnerPanel = (props: Props) => {
...
+            </TwitterShareButton>
+          </div>
+
+          <div className={styles.socialIconContainer}>
+            <EmailShareButton url={location} className="share">
+              <EmailIcon size={48} round={true} />
+            </EmailShareButton>
+          </div>
+        </div>
+      }
```
bvillaroman, line 209: it looks like you're the most familiar with the new Owner Panel. So soon we should convert the styles to styled-components. In the next couple of prs you should see how i do styled components for components that apply for just that page. If we find ourselves reusing code then we can do another type of style of creating styled-components. But this is great!

danthaman44: Sounds good!

PR #103, opened 2020-05-08
Add email validation to input by using simple regex

Yong-L:
This invalidates emails of example@example.

The following should be valid
* example@example.com
* example@example.in.edu


Yong-L:
Should fix

* https://trello.com/c/QHWdQNnP/228-entering-a-valid-email-after-entering-an-invalid-one-doesnt-let-you-submit
* https://trello.com/c/4PYT2txM/229-use-regex-to-prevent-people-from-entering-an-invalid-email

PR #108, opened 2020-05-09
Added navigation header to merchant page

danthaman44:
Screen Shot 2020-05-09 at 7 52 42 PM

Screen Shot 2020-05-09 at 7 55 23 PM


danthaman44:
The NavBar doesn't match the latest designs yet, but @bvillaroman is working on that

bvillaroman:
So i was already working on this and just forgot to say it in the trello, but i was going to make it so that footers and headers are on all pages of the app so we don't have to manually import them in each page. Sorry i haven't said anything earlier, but i should be able to finish that by tonight

danthaman44:
@bvillaroman Ahh got it. Automatically including the headers sounds like the right approach to me, thanks for the heads up

PR #110, opened 2020-05-11
Quickfix for numsupporters tooltip on merchants page

Jeremy-Xue:
Left a table style in the wrong .scss file, tooltip is misaligned in merchants directory page

### DIFF
--- a/src/components/ProgressBar/styles.module.scss
+++ b/src/components/ProgressBar/styles.module.scss
@@ -20,6 +20,7 @@
   justify-content: space-between;
 }
 
-table {
+.tooltipTable {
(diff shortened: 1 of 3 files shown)


src/components/ProgressBar/styles.module.scss
```diff
@@ -19,3 +19,7 @@
   display: flex;
   justify-content: space-between;
 }
+
+table {
```
jtmckibb, line 23: In general it's better to give it a class and give the style to that class. Otherwise the next person that adds a table to this file will see a style that doesn't make sense.

PR #112, opened 2020-05-13
Bruce/add header footer for all pages rebase2

bvillaroman:
combining the header fix branch with the add header footer to all pages branch and rebased


src/components/App/App.tsx
```diff
@@ -18,41 +20,47 @@ history.listen((location) => {
...
-const AboutPage = lazy(() => import('../About'));
 const MerchantsPage = lazy(() => import('../MerchantsPage'));
 const ErrorPage = lazy(() => import('../404Page'));
 
-class App extends React.Component<{}> {
-  render() {
+const App = () => {
+  const [menuOpen, setMenuOpen] = useState(false);
+
+  const returnComponent = (child) => {
```
danthaman44, line 29: Siiiick

src/components/Navbar/NavBar.tsx
```diff
@@ -189,21 +195,33 @@ const LanguageContainer = styled.div`
...
+    color: #a7182d;
+  }
 `;
 const ReactNavLink = styled(Link)`
   text-decoration: none;
   color: black;
+  transition: 0.2s;
   ${(props: CompactProps) =>
-    props.compact &&
+    props.compact === 'true' &&
```
danthaman44, line 215: any reason is a string and not a boolean?

bvillaroman: chrome throws a console warning saying it should be a string if youre using props to change css through styled components

PR #113, opened 2020-05-15
Athena/merchants page

ath-chen:
- made donation pool & fixed square modal css

to do: 
- fix css throughout merchants page to be styled components

Screen Shot 2020-05-14 at 8 46 51 PM

Screen Shot 2020-05-14 at 9 06 39 PM


src/components/MerchantsPage/DonationPool.tsx
```diff
@@ -0,0 +1,80 @@
...
+          <h4>
+            Checked out our merchants but not sure who to donate to first?
+          </h4>
+          <span>You can support by donating to our donation pool!</span>
+          <span>
+            All donations will be distributed evenly to our merchants.
+          </span>{' '}
+          <br />
+          <button className="button--red-filled" onClick={openModal}>
+            SUPPORT CHINATOWN
```
juliexxia, line 36: Thoughts on making the button text something like "MAKE A POOL DONATION" or something that is more direct about what action you're taking when you click the button?

src/components/MerchantsPage/DonationPool.tsx
```diff
@@ -0,0 +1,80 @@
...
+  return (
+    <React.Fragment>
+      <Container>
+        <ColumnContainer>
+          <h4>
+            Checked out our merchants but not sure who to donate to first?
+          </h4>
+          <span>You can support by donating to our donation pool!</span>
+          <span>
+            All donations will be distributed evenly to our merchants.
```
juliexxia, line 32: So technically it'll only get distributed to the merchants who accept donations, which so far is all of them but might change in the future? Idk how important it is to be clear about that here. Maybe it'll just muddy up the text.

src/components/MerchantsPage/MerchantsPage.tsx
```diff
@@ -93,6 +95,11 @@ const MerchantsPage = (props: Props) => {
               </div>
             </div>
           </div>
+
+          <ModalPaymentProvider>
+            <DontaionPool />
```
juliexxia, line 100: typo: DonationPool?

src/components/MerchantsPage/DonationPool.tsx
jtmckibb: Might just muddy up our text. I think it's fine for now, we can think about adding a disclaimer for legal reasons in the future

src/components/MerchantsPage/DonationPool.tsx
jtmckibb: cc @nanxiy but I do like the smaller text

src/components/MerchantsPage/MerchantsPage.tsx
ath-chen: good catch!

src/components/MerchantsPage/DonationPool.tsx
nanxiy: “Donate to Chinatown”? Idk something shorter is better esp for mobile

PR #127, opened 2020-05-20
Adding link to download flyer and fixing vertical spacing on donation modal

jacoblb64:
- Added flyer asset (for english only)
- Added a small new flyer section at the bottom of the merchant page with some basic styling
- Reduced height of donation modal from 80% to 50% for larger than tablet views

Card links:
- [Adding the Flyer Link](https://trello.com/c/0OiBnQ0h/389-add-download-flyers-link-at-bottom-of-merchant-directory-page)
- [Fixing the padding for the donation modal](https://trello.com/c/5hCEXpsr/421-padding-for-the-donation-vouchers-pop-up-is-weird)

---
Screenshots:

Flyer link (shown on mobile):
Screen Shot 2020-05-19 at 10 00 59 PM

Donation modal:
Screen Shot 2020-05-19 at 9 54 53 PM


src/components/MerchantsPage/MerchantsPage.tsx
```diff
@@ -112,6 +113,12 @@ const MerchantsPage = (props: Props) => {
               </div>
             </div>
           </MerchantInfoContainer>
+
+          <FlyerContainer>
+            <p>
+              Know of any business owners that fit our target merchant? <a download href={sclFlyerEnglish}>Download our flyer to share with them.</a>
```
jtmckibb, line 119: These should be translated using t(...)

src/components/MerchantsPage/MerchantsPage.tsx
```diff
@@ -7,6 +7,7 @@ import DescriptionBox from './DescriptionBox';
 import ContributionBar from './ContributionBar';
 import styles from './styles.module.scss';
 import nycMapBackground from './images/nyc_3.png';
+import sclFlyerEnglish from './images/scl-flyer-english.png'
```
jtmckibb, line 10: @nanxiy should we also include the Chinese version here? Maybe an English speaker donor knows a Chinese merchant and might not know the click the Chinese toggle at the top

src/components/MerchantsPage/MerchantsPage.tsx
```diff
@@ -130,3 +137,18 @@ const MerchantInfoContainer = styled.div`
     flex-direction: column-reverse;
   }
 `;
+
+const FlyerContainer = styled.div`
+  margin: 3vw 7vw;
+  font-weight: bold;
+  
+  a {
```
jtmckibb, line 145: I personally prefer to define classes instead of styling tags, but seems like we have no standard here

src/components/MerchantsPage/MerchantsPage.tsx
```diff
@@ -130,3 +137,18 @@ const MerchantInfoContainer = styled.div`
...
+    color: #a7182d;
```
jtmckibb, line 146: Are you sure this color isn't defined in a shared var somewhere?

jtmckibb: I just checked and it's not. Maybe we should?

jtmckibb:
Thanks for fixing the donation modal height!

src/components/MerchantsPage/MerchantsPage.tsx
jacoblb64: Yeah so it looks like we do have `$standard-red` that's close (I actually pulled this color off of the 404 page, which is slightly different!), which is defined in `index.scss`, but that doesn't' seem to be referenced outside of that file at all

jacoblb64: I can see if I can refactor to reference that / maybe put it in a shared place

jtmckibb: I wouldn't say it's blocking. @nanxiy any reason why we are using two different types of reds here?

jtmckibb: But yeah in general it seems like all of our links are going to be this color red, so if we can move this out and make all of our hyperlinks this color that would be great

src/components/MerchantsPage/MerchantsPage.tsx
jacoblb64: Ah thanks for the catch, was wondering how the language toggle worked! Let me add that

src/components/MerchantsPage/MerchantsPage.tsx
jacoblb64: Either way works for me, I can add a class to be more specific

src/components/MerchantsPage/MerchantsPage.tsx
jtmckibb: The language toggle doesn't currently work because we haven't input the Chinese translations yet, but we should stay on top of supporting it

src/components/MerchantsPage/MerchantsPage.tsx
jtmckibb: In the case that we make all of our links look like this, then it can be the a tag. But if it's just isolated to this one section, then I'd say use a class. Now thinking of it, we should probably  use the tag and pull it out into a shared style

src/components/MerchantsPage/MerchantsPage.tsx
jacoblb64: Will followup with the translation support and color refactoring in another MR 👍

PR #131, opened 2020-05-21
Refactor merchant page styling and using t() and now using flyer full zip

jacoblb64:
- Refactored styling of the base MerchantsPage to all draw from local styles.module.scss
- Adding useTranslation() to all text on the page
- Added a redLink style that can be reused, and which reuses $standard-red from index.scss
- Replacing the english png download with the zip, hosted in the public folder


danthaman44:
Niiiiice! Ideally we can add our localized strings to `src/locales/en/translation.json`, but that be a separate PR

src/components/MerchantsPage/MerchantsPage.tsx
```diff
@@ -98,30 +104,30 @@ const MerchantsPage = (props: Props) => {
...
 
             <div className={styles.storeInfoContainer}>
               <NavBar filterStoreType={filterStoreType} />
 
               <div className={styles.merchantsContainer}>
                 {filter.map((store: any) =>
-                  store!.seller_id !== 'send-chinatown-love' ? (
-                    <MerchantCard key={store!.seller_id} storeInfo={store} />
+                  store.seller_id !== 'send-chinatown-love' ? (
+                    <MerchantCard key={store.seller_id} storeInfo={store} />
```
jacoblb64, line 116: I'm pretty sure these null checks were extraneous as `.map()` will only return non-null items. Someone feel free to correct me if I'm missing context.

jacoblb64:
> Niiiiice! Ideally we can add our localized strings to `src/locales/en/translation.json`, but that be a separate PR

Gotchya, thanks for the heads up! Would we wait to have both cn and en strings before updating that?

jtmckibb:
we just fill them both with english for now. Ling is working on translating
them now it’s just taking some time

src/components/MerchantsPage/MerchantsPage.tsx
```diff
@@ -76,13 +78,17 @@ const MerchantsPage = (props: Props) => {
                 <h2 style={{ fontWeight: 'bolder' }}>Our Chinatown</h2>
                 <br />
                 <p>
-                  We are providing an online platform to low-tech, cash-only,
-                  Asian-owned small businesses that have been disproportionately
-                  impacted by COVID-19.
+                  {t(`
```
jtmckibb, line 81: This should have the key, and the value should be the string itself

jacoblb64: ...that makes a lot more sense ahha, I was following what was done in the DonationPool component, but I guess that wasn't set up properly either. I refactored both components' text to `translation.json`

jtmckibb:
LGTM!

PR #132, opened 2020-05-21
Added underline to Merchant navlink to mimic Weebly

danthaman44:
Screen Shot 2020-05-20 at 11 26 30 PM


danthaman44:
PR looks weird, but its mostly just autoformatting

PR #133, opened 2020-05-22
Added copy to clipboard button

danthaman44:
Screen Shot 2020-05-21 at 11 28 29 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,16516 @@
+{
+  "name": "sendchinatownlove",
+  "version": "0.1.0",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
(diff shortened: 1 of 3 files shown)


jtmckibb:
Nit but the graphic doesn't look centered in the button. Maybe there is some whitespace on the left of the graphic that needs to be trimmed? @nanxiy do you mind helping with the trimming?

nanxiy:
I have new icons for the social share icons on figma if you want to grab those!

jtmckibb:
Actually let's leave it as is then since this is going to be replaced next week with the new social share tab

danthaman44:
ahh yeah the SVG was larger than the button itself so I had add some padding to shrink + center it...this was the best I could do with my mediocre CSS skills 😅

Issue #141, opened 2020-05-23
Allow people to make more than one donation to the same merchant without having to refresh the page


jtmckibb:
https://trello.com/c/EhA1MzcF/428-uuid-for-submitting-a-payment-is-bound-to-the-global-state-instead-of-the-localized-modal-state

Issue #140, opened 2020-05-23
Update pooled donation copy in payment flow


jtmckibb:
https://trello.com/c/YF6bdPLL/426-add-language-to-the-donation-pool-payment-flow

PR #148, opened 2020-05-25
fix for #140, #141

TaskSquirrel:
Fixes for #140 and #141 
Plus some styling changes to match spec
* transparent background on inactive filter buttons on merchants page
* flyer message left aligned

Before:
image

After:
image

### DIFF
--- a/src/components/MerchantsPage/styles.module.scss
+++ b/src/components/MerchantsPage/styles.module.scss
@@ -145,8 +145,17 @@
 }
 
 .flyerContainer {
-  margin: 3vh 7vw;
+  $flyer-margin: 3rem;
+
+  margin-top: $flyer-margin;
+  margin-bottom: $flyer-margin;
+  font-size: 120%;
   font-weight: bold;
+
+  @include for-phone-only {
+    margin-left: $flyer-margin;
+    margin-right: $flyer-margin;
+  }
 }
 
 a.redLink {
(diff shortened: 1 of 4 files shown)


danthaman44:
Can you upload a screenshot of the UI after these changes?

src/components/MerchantsPage/styles.module.scss
```diff
@@ -145,8 +145,17 @@
 }
 
 .flyerContainer {
-  margin: 3vh 7vw;
+  $flyer-margin: 3rem;
+
+  margin-top: $flyer-margin;
+  margin-bottom: $flyer-margin;
+  font-size: 120%;
```
jtmckibb, line 152: Why is the font size a percent?

src/index.scss
```diff
@@ -107,6 +107,7 @@ h3 {
 
 .modalButton {
   &--back {
+    background: none;
```
jtmckibb, line 110: What does this fix?

TaskSquirrel: Removes background color from the filtering buttons on the top of the screenshots in Chrome; appears to be using the Chrome default button bg color

TaskSquirrel:
Added before & after (top filtering buttons and bottom flyer message)

src/components/MerchantsPage/styles.module.scss
TaskSquirrel: The text in the flyer msg seems to be slightly larger than the normal text; is there any text component/mechanism you guys are using to keep font sizes consistent?

Issue #153, opened 2020-05-26
Round to the nearest dollar


jtmckibb:
https://trello.com/c/W85YwhLc/455-update-the-frontend-to-round-to-nearest-dollar

Issue #150, opened 2020-05-25
Add max width for owner panel


jtmckibb:
https://trello.com/c/7QHR5tOc/453-max-width-for-owner-panel

Issue #139, opened 2020-05-23
Update fonts of payment flow buttons


jtmckibb:
https://trello.com/c/7QGoWyFd/422-update-button-fonts-for-back-and-submit-next-buttons-in-payment-flow

PR #154, opened 2020-05-26
Fixes for #153, #150, #139

TaskSquirrel:
#153 Floors to nearest dollar
Before:
image

After:
image

#150 Cap owner panel width at 40% of container - should it be a little smaller?
image

#139 Fix "modal button" font family and casing
Before:
image

After:
image


src/components/SellerPage/SellerPage.tsx
```diff
@@ -90,7 +90,7 @@ const ContentContainer = styled.div`
   max-width: 1200px;
   min-height: 1200px;
   @media (min-width: 900px) {
-    grid-template-columns: 1fr minmax(405px, 1fr);
+    grid-template-columns: 1fr minmax(405px, 40%);
```
jtmckibb, line 93: I'm not sure we want to max it at a percentage — I think we want the max width to be the max width it has to be in order to support the width of two buttons side by side

But it should be a pixel length

jtmckibb: My thinking is that: 

max_width = however wide it is on Figma in px

For mobile, the owner panel will be full width
For desktop, it would be max_width until it takes over more than 50% of the screen, then it would switch to mobile

TaskSquirrel: Sounds like the owner panel should just be fixed at 464px? Sinec it's the panel's gonna span the screeen above 900px screen size anyway

jtmckibb: ah yeah it should be fixed

jtmckibb: @bvillaroman how come we decided to update this to a percent? It should be fixed at 464px, and if it takes up more than 50% of the screen, then we should convert to mobile view no?

src/components/SellerPage/SellerPage.tsx
```diff
@@ -90,7 +90,7 @@ const ContentContainer = styled.div`
   max-width: 1200px;
   min-height: 1200px;
   @media (min-width: 900px) {
-    grid-template-columns: 1fr minmax(405px, 40%);
+    grid-template-columns: 1fr 464px;
```
TaskSquirrel, line 93: "fixed" it 😃

jtmckibb: ahaha nice XD

PR #157, opened 2020-05-29
Create multiple line items instead of aggregating the total sum

Yong-L:


src/components/ModalPayment/SquareModal.tsx
```diff
@@ -63,13 +69,23 @@ const SquareModal = ({
       return;
     }
 
+    console.log(numberOfMeals);
```
stanleyzhengnyc, line 72: theres an extra console log here

Yong-L: good catch

PR #158, opened 2020-05-29
Added Facebook tracking for user interactions

danthaman44:
Test plan:
-Verified tracking events are being sent via FB Pixel Helper Chrome extension
-Verified tracking events show up in Ads Manager
Screen Shot 2020-05-28 at 10 40 24 PM
Screen Shot 2020-05-28 at 10 09 29 PM

### DIFF
--- a/src/components/App/App.tsx
+++ b/src/components/App/App.tsx
@@ -6,6 +6,7 @@ import Loader from '../Loader';
 import Header from '../Navbar';
 import Footer from '../Footer';
 import { ModalPaymentProvider } from '../../utilities/hooks/ModalPaymentContext/context';
+import ReactPixel from 'react-facebook-pixel';
 
 const trackingId = process.env.REACT_APP_API_ENDPOINT!;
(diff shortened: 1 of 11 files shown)


jtmckibb:
wauuu

src/components/App/App.tsx
```diff
@@ -29,6 +30,13 @@ const SellerPage = lazy(() => import('../SellerPage'));
 const MerchantsPage = lazy(() => import('../MerchantsPage'));
 const ErrorPage = lazy(() => import('../404Page'));
 
+const options = {
+  autoConfig: true, // set pixel's autoConfig
+  debug: true, // enable logs
+};
+const pixelId = '781939739005509';
```
jtmckibb, line 37: This isn't private right?

danthaman44: ahh shoot thats a good call, this needs to be private

danthaman44: Ok I changed it to be an env variable. Its probably best that we only set the env variable in prod, so we don't fire tracking events from local or staging.

PR #168, opened 2020-05-30
Bruce/adding gift card redemption rebase2

bvillaroman:
lol prepare yall selves this was alot


src/utilities/api/interactionManager.ts
```diff
@@ -105,3 +105,15 @@ export const makeSquarePayment = async (
       throw err;
     });
 };
+
+export const getVoucher = async (id: string) =>
+  axios
+    .get(vouchers + id)
+    .then((res) => res)
+    .catch((err) => err);
```
jtmckibb, line 113: Are errors caught/surfaced to the client?

src/pages/VoucherRedemption/StoreBanner.tsx
```diff
@@ -0,0 +1,70 @@
...
+  ${(props: ContainerProps) => !props.isOnLandingPage && 'margin-left: 15px;'};
+  font-style: normal;
+  font-weight: bold;
+  font-size: 22px;
+  line-height: 30px;
+`;
+const OwnerImage = styled.img`
+  width: ${(props: ContainerProps) =>
+    props.isOnLandingPage ? '150px' : '55px'};
+  // height: ${(props: ContainerProps) =>
```
jtmckibb, line 64: Is this intended to be commented out?

src/utilities/api/interactionManager.ts
jacoblb64: I think we can switch these to await calls instead of .then() and if there any errors they'll be thrown up the regular call stack instead of caught here--then maybe in the component catch the error and display a message?

src/pages/VoucherRedemption/VoucherRedemption.tsx
```diff
@@ -0,0 +1,106 @@
...
+          {/* {view === 0 && <Backdrop src={NYCBackdrop}/>} */}
+        </>
+      )}
+    </Container>
+  );
+};
+
+export default VoucherRedemption;
+
+const Container = styled.div`
```
jacoblb64, line 90: why not have this in `styles.ts`?

src/pages/VoucherRedemption/VoucherRedemption.tsx
```diff
@@ -0,0 +1,106 @@
...
+      dispatch({ type: SET_VOUCHER_INFO, payload: voucher });
+      setLoading(false);
+    } catch {
+      params.push('/');
+      setLoading(false);
+    }
+  };
+
+  useEffect(() => {
+    setLoading(true);
```
jacoblb64, line 55: any reason to have this in `useEffect()` directly, and not within `fetchData()`? I don't see any issues with this, but could be cleaner to have `fetchData()` be the exclusive changer of loading state.

src/pages/VoucherRedemption/Landing.tsx
```diff
@@ -0,0 +1,123 @@
...
+      <FooterContainer>
+        <Image src={Logo} />
+      </FooterContainer>
+    </Container>
+  );
+};
+
+export default LandingCard;
+
+const Container = styled.div`
```
jacoblb64, line 53: same comment about `styles.ts` as before

src/components/App/App.tsx
```diff
@@ -73,9 +77,14 @@ const App = () => {
...
           <Route path="/all">{returnComponent('all')}</Route>
+          <Route path="/voucher/:id">
+            <VoucherProvider>
+              <VoucherRedemptionPage />
+            </VoucherProvider>
+          </Route>
           <Route path="/merchants">{returnComponent('merchants')}</Route>
           <Route path="/:id">{returnComponent('seller')}</Route>
-          <Route path="/:id#story">{returnComponent('seller')}</Route>
+          <Route path="/:id#story">{returnComponent('seller')}</Route> */}
```
jacoblb64, line 87: what's this `*/`? is that leftover from commenting something out?

src/utilities/api/interactionManager.ts
bvillaroman: some of these api errors are caught and we either show the api error or make our own

src/pages/VoucherRedemption/StoreBanner.tsx
bvillaroman: hmmm i have to look at it again

src/pages/VoucherRedemption/VoucherRedemption.tsx
bvillaroman: the container was specific to this page, but we can modularize it like how i've done with the other components

src/pages/VoucherRedemption/VoucherRedemption.tsx
bvillaroman: i wanted to load the component with the loader before we fetched any data.

src/components/App/App.tsx
bvillaroman: this might have gotten added during some merge conflicts i was trying to resolve

PR #170, opened 2020-05-30
Static and Dynamic Translation for Merchant and Seller pages

jacoblb64:
- Updated all of our static (compile-time) translation content with the latest (5/19) translations from the google doc "webpage final review"
- Watching `i18n.language` to trigger `useEffects()` which in turn re-triggers server requests with the new locale
- Will make requests for the rest of the content that needs to be translated through the google form

**Could we have someone who speaks Chinese confirm that everything here makes sense?** 

Attaching screenshots below:

Screen Shot 2020-05-30 at 1 03 37 PM
Screen Shot 2020-05-30 at 1 03 53 PM


stzhng:
this is great @jacoblb64 !

src/components/MerchantsPage/ContributionBar.tsx
```diff
@@ -1,20 +1,23 @@
...
 import styled from 'styled-components';
+import { useTranslation } from 'react-i18next';
 
 interface Props {
   totalDonations: number;
   totalGiftCards: number;
 }
 
 const ContributionBar = ({ totalDonations, totalGiftCards }: Props) => {
+  const { t } = useTranslation();
```
danthaman44, line 11: I wonder if its possible to declare { t } in App.tsx so that its available to the whole app by default...(just an idea, may or may not be possible)

jacoblb64: I'm fairly new to React, but I think the only ways to achieve something like this would be to pass it from the App component down through to child components, or to set it as context. I don't think either are cleaner than this though--what do you think?

jacoblb64:
After talking with @jtmckibb, going to go ahead and merge this now, then reach out to our community to verify translated text on the live site. We'll fix any typos / corrections in a followup MR.

PR #178, opened 2020-06-10
Single Use Voucher Redemptions and Overall Redemption Flow Tweaks

jacoblb64:
New:
- Added checks for single_use variable and modifying the pages and flow when this is true
- Various text changes when in the single_use flow

Changes to both flows:
- Improved error handling on the amount input box
  - Accepts inputs between 5¢ and the full amount
  - Error state now disables and greys out the "Next" Button
  - Error state changes the border to red
- Added floating $ sign to input box
- Fixed logic on when to invert the colors for the More Info tool tip (code is intentionally verbose to be more readable)
- Flexbox changes to allow spacing to be added more naturally on different size devices
- Changed the back / cancel buttons to more closely match the Figma design
- Removed the z-index the More Info div which was blocking the input box when not popped out

General code cleanup:
- Fixed some spelling errors
- Removed rogue end comment tag
- Removed a rogue ;

Verified amount changes persist to the db ✓

---

Screenshots of both flows: 
 - Shown on iPhone X size
 - Other states shown in multi-flow work for single-flow as well)

Multi flow:
Redemption Multi Flow

Single flow:
Redemption Single Flow

### DIFF
--- a/src/pages/VoucherRedemption/Amount.tsx
+++ b/src/pages/VoucherRedemption/Amount.tsx
@@ -12,7 +12,7 @@ import {
 } from '../../utilities/hooks/VoucherContext/constants';
 import {
   AmountContainer,
-  MessageConatiner,
+  MessageContainer,
   Text,
   Footer,
   NextButton,
@@ -81,7 +81,7 @@ const Amount = (props: Props) => {
           max="10000"
         />
       </AmountContainer>
-      <MessageConatiner>
+      <MessageContainer>
         <Text size="16px" color="#DD678A" width="50%">
           {error}
         </Text>
@@ -93,7 +93,7 @@ const Amount = (props: Props) => {
             Remaining
           </Text>
         </Text>
-      </MessageConatiner>
+      </MessageContainer>
       <Footer>
         <Voucher>
           Voucher Code: <Bold>{voucher.seller_gift_card_id}</Bold>{' '}
--- a/src/pages/VoucherRedemption/Amount.tsx
+++ b/src/pages/VoucherRedemption/Amount.tsx
@@ -18,6 +18,7 @@ import {
   NextButton,
   Voucher,
   Bold,
+  FlexFillSpace,
 } from './styles';
 
 interface Props {}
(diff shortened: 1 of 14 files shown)


src/pages/VoucherRedemption/Amount.tsx
```diff
@@ -38,7 +43,7 @@ const Amount = (props: Props) => {
 
   const handleAmount = (e) => {
     setError('');
-    if (e.target.value * 100 > voucher.amount)
+    if (e.target.value * 100 > voucher.amount || e.target.value * 100 < 5)
```
jtmckibb, line 46: Why 5 cents?

jacoblb64: Honestly up to you! I saw @bvillaroman had a minValue of 5 (cents) before, so I left it at that. This could be 1¢, $1, or whatever we want

jtmckibb:
1 cent seems fine

src/pages/VoucherRedemption/Amount.tsx
jacoblb64: Changed this to 1 cent!

src/pages/VoucherRedemption/MoreInfo.tsx
```diff
@@ -43,6 +47,29 @@ const LandingCard = (props: Props) => {
...
 
+const getColor = (
+  foreground: Boolean,
+  showInfo?: Boolean,
+  inverted?: Boolean
+): string => {
+  const white = 'white';
+  const red = '#ab192e';
+
+  if (!inverted) {
```
danthaman44, line 58: very minor nit, I think you can reverse the logic here to make the logic cleaner
```
if (inverted) {
 // code
} else {
 // code
}

File: src/pages/VoucherRedemption/MoreInfo.tsx [hunk 3]
```diff
@@ -43,6 +47,29 @@ const LandingCard = (props: Props) => {
...
+    if (foreground) {
+      return !showInfo ? red : white;
+    } else {
+      return !showInfo ? white : red;
+    }
+  } else {
+    if (foreground) {
```
danthaman44, line 65: consider ternary syntax here?

src/pages/VoucherRedemption/MoreInfo.tsx
jacoblb64: To this and the other comment--so I was trying to spell this out to be as readable as possible, so I kept the negative case on top, as it's the "default" case, and didn't use a second ternary.

That being said, if you think it'd be more readable the other way, than that seems like a sign I should change it!

src/pages/VoucherRedemption/Confirm.tsx
```diff
@@ -62,54 +65,80 @@ const Amount = (props: Props) => {
...
+        <MoreInfo marginLeft="35px" showShadow={true} inverted={true} />
       </Header>
-      <MessageConatiner>
+      {!voucher.single_use ? (
+        <MessageContainer>
+          <Text size="16px" width="50%">
+            Voucher balance
+          </Text>
+          <Text size="16px" width="50%" align="flex-end">
+            ${(voucher.amount / 100).toFixed(2)}
```
danthaman44, line 104: maybe extract `(voucher.amount / 100).toFixed(2)` into a `const` since its used in multiple places

src/pages/VoucherRedemption/MoreInfo.tsx
danthaman44: nah totally up you! Just a matter of preference :)

src/pages/VoucherRedemption/Confirm.tsx
jacoblb64: I realized it's actually only in two places, and it's a slightly different calculation (one's the amount, one's the remainder). But I think there could be room for a better conversion system from simple decimals to the way we store it in our database as an int * 100

PR #180, opened 2020-06-13
Fixed critical bug with location rendering in the redemption flow

danthaman44:
-This fixes a bug which causes the redemption flow to break because it can't render the location from the backend correctly.
-Also fixes a minor bug where the `color` prop in the `Button` component wasn't being set properly due to a missing semicolon

Before:
Screen Shot 2020-06-13 at 3 55 47 PM

After:
Screen Shot 2020-06-13 at 4 10 13 PM

### DIFF
--- a/src/pages/VoucherRedemption/Landing.tsx
+++ b/src/pages/VoucherRedemption/Landing.tsx
@@ -52,7 +52,11 @@ const LandingCard = (props: Props) => {
         <br />
       </CardContainer>
       <br />
-      <Button color="#ab192e">{voucher.locations[0]}</Button>
+      <Button color="#ab192e">
+        {voucher.location.line1}
+        <br></br>
+        {voucher.location.line2}
+      </Button>
       <FooterContainer>
         <Image src={Logo} />
       </FooterContainer>
@@ -107,9 +111,9 @@ const SupportingText = styled.span`
 `;
 const Button = styled.div`
   cursor: pointer;
-  ${(props: ButtonProps) => props.color && `color: ${props.color}`}
+  ${(props: ButtonProps) => `color: ${props.color ?? ''};`}
   font-size: 13px;
-  line-height: 18px;
+  line-height: 1.5;
   text-align: center;
 `;
 const Balance = styled.div`
--- a/src/pages/VoucherRedemption/VoucherRedemption.tsx
+++ b/src/pages/VoucherRedemption/VoucherRedemption.tsx
@@ -15,6 +15,7 @@ import { SET_VOUCHER_INFO } from '../../utilities/hooks/VoucherContext/constants
 import Loader from '../../components/Loader';
 import NYCBackdrop from '../../images/nyc-background.png';
 import { getVoucher, getSeller } from '../../utilities/api/interactionManager';
+import { getLocationInfo } from '../../utilities/hooks/helpers';
 
 interface Props {}
 interface ContainerProps {
@@ -40,7 +41,7 @@ const VoucherRedemption = (props: Props) => {
         ownerImage: merchantData.data.owner_image_url,
         storeImage: merchantData.data.hero_image_url,
         sellerID: seller_id,
-        locations: merchantData.data.locations,
+        location: getLocationInfo(merchantData),
       };
 
       dispatch({ type: SET_VOUCHER_INFO, payload: voucher });
--- a/src/pages/VoucherRedemption/Confirm.tsx
+++ b/src/pages/VoucherRedemption/Confirm.tsx
@@ -26,6 +26,7 @@ import {
   Bold,
   FlexFillSpace,
 } from './styles';
+import { getLocationInfo } from '../../utilities/hooks/helpers';
 
 interface Props {}
 interface ContainerProps {
@@ -58,7 +59,7 @@ const Amount = (props: Props) => {
         ownerImage: merchantData.data.owner_image_url,
         storeImage: merchantData.data.hero_image_url,
         sellerID: seller_id,
-        locations: merchantData.data.locations,
+        location: getLocationInfo(merchantData),
       };
 
       dispatch({ type: SET_VOUCHER_INFO, payload: newVoucher });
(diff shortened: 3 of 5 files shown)


src/pages/VoucherRedemption/Landing.tsx
```diff
@@ -40,7 +40,12 @@ const LandingCard = (props: Props) => {
         <br />
       </CardContainer>
       <br />
-      <Button color="#ab192e">{voucher.locations[0]}</Button>;
+      <Button color='#ab192e'>
+        {voucher.location.line1}
```
jtmckibb, line 44: are you sure that we only have a single location?

src/pages/VoucherRedemption/VoucherRedemption.tsx
```diff
@@ -28,6 +29,34 @@ const VoucherRedemption = (props: Props) => {
...
+    let getLocationInfo = (merchantData): LocationInfo => {
+      const emptyLocationInfo = {
+        line1: '',
+        line2: '',
+      };
+      if (!merchantData || !merchantData.data) {
+        return emptyLocationInfo;
+      }
+      const locations = merchantData.data.locations;
+      if (!locations || !locations.length) {
```
jtmckibb, line 41: nit: but I prefer locations.length > 0 to not rely on the truth values of .length

src/pages/VoucherRedemption/VoucherRedemption.tsx
```diff
@@ -28,6 +29,34 @@ const VoucherRedemption = (props: Props) => {
...
+        return emptyLocationInfo;
+      }
+      const location = locations[0];
+      if (!location) {
+        return emptyLocationInfo;
+      }
+      const locationInfo = {
+        line1: (location.address1 ?? '') + ' ' + (location.address2 ?? ''),
```
jtmckibb, line 49: Could you use JavaScript string interpolation here?

src/pages/VoucherRedemption/VoucherRedemption.tsx
```diff
@@ -28,6 +29,34 @@ const VoucherRedemption = (props: Props) => {
...
+        line2:
+          (location.city ?? '') +
```
jtmckibb, line 51: Same

src/pages/VoucherRedemption/VoucherRedemption.tsx
danthaman44: Works for me!

src/pages/VoucherRedemption/Landing.tsx
danthaman44: I thought about this a bit. All of our merchants right now only have one location, and a lot of the UI in the sellers page and redemption flow is predicated on there only being one location. Given our target merchant demographic (mom and pop stores), I think its safe to assume that there will only be one location for now. If that changes in the future, we can re-evaluate the product experience.

src/pages/VoucherRedemption/VoucherRedemption.tsx
danthaman44: Good call!

src/pages/VoucherRedemption/Landing.tsx
```diff
@@ -40,7 +40,12 @@ const LandingCard = (props: Props) => {
...
+        <br></br>
+        {voucher.location.line2}
+      </Button>
+      ;
```
jacoblb64, line 48: watch out with this `;`, I think it'll be rendered in the UI

src/pages/VoucherRedemption/Landing.tsx
```diff
@@ -95,9 +100,9 @@ const SupportingText = styled.span`
 `;
 const Button = styled.div`
   cursor: pointer;
-  ${(props: ButtonProps) => props.color && `color: ${props.color}`}
+  ${(props: ButtonProps) => props.color && `color: ${props.color};`}
```
jacoblb64, line 103: unrelated to your change, but does this line just add `undefined` when props.color is undefined? Maybe we should do a ternary operator with `''` if falsy

src/pages/VoucherRedemption/VoucherRedemption.tsx
```diff
@@ -28,6 +29,29 @@ const VoucherRedemption = (props: Props) => {
   const [loading, setLoading] = useState(false);
 
   const fetchData = async () => {
+    let getLocationInfo = (merchantData): LocationInfo => {
+      const emptyLocationInfo = {
+        line1: '',
+        line2: '',
+      };
+      if (!merchantData || !merchantData.data) {
```
jacoblb64, line 37: You might be able to use [Optional Chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) here to make this a little bit cleaner if you want. Could combine all of these checks below into one or a few statements.

src/pages/VoucherRedemption/Landing.tsx
jtmckibb: I agree here. I think we should amend the API to only return a single location, and in parallel make the frontend handle only that location as well

jtmckibb: https://github.com/sendchinatownlove/ruby/issues/157

jtmckibb: https://github.com/sendchinatownlove/sendchinatownlove.github.io/issues/185

src/pages/VoucherRedemption/Landing.tsx
danthaman44: Good catch!

src/pages/VoucherRedemption/Landing.tsx
danthaman44: Cool I got it to work using the null coalesce operator
`${(props: ButtonProps) => `color: ${props.color ?? ''};`}`

src/pages/VoucherRedemption/VoucherRedemption.tsx
danthaman44: ooh nice! Did not know about this

Issue #175, opened 2020-06-04
Disable confirm button when payment's submitted


TaskSquirrel:
The confirm button in the square payment submission modal should be disabled once the user clicks it to prevent multiple submissions on the front-end side

![image](https://user-images.githubusercontent.com/19297741/83765671-7725a280-a649-11ea-8c78-0796a5088ed0.png)

PR #188, opened 2020-06-15
Disable payment confirm button after form submission

yanhao-li:
This PR fixes https://github.com/sendchinatownlove/sendchinatownlove.github.io/issues/175
- Disabled submit button after payment submission, Square will also block duplicate payment on their side though, thus this issue should not be a priority.
- Changed the disabled button cursor style to default.

Before (I'm keeping clicking the submit button, and you could see the request do have been sent multiple times):
before

After:
after


justintmckibben:
Thanks for doing this!

PR #190, opened 2020-06-16
changed sizing and hide tabs if empty

ath-chen:


src/components/StoreInfo/StoreInfoContainer.tsx
```diff
@@ -27,7 +27,11 @@ export const StoreInfo: React.SFC<Props> = ({ seller }) => {
   };
 
   // logic for nav bar & tab switching
-  const storeNavItems = ['story', 'menu', 'gallery'];
+  const storeNavItems = [
+    story && 'story',
+    seller.menu_url && 'menu',
+    seller.gallery_image_urls.length > 0 && 'gallery',
```
justintmckibben, line 33: i'm confused by this — wouldn't it just always return the truthy values of the left param? JavaScript antics

ath-chen: ya! so if they're empty instead of having 'story', 'menu', etc it'll just be null. We're doing a check on line 85 which would then display the tab or not

PR #195, opened 2020-06-17
quick fix to voucher page

bvillaroman:
- temp fix to have components stacked on top of each other instead of overlapping


justintmckibben:
TY this looks great!

PR #207, opened 2020-06-27
Delegating GAM Payments to Non-Profit Square accounts (Think!Chinatown)

jacoblb64:
To fully facilitate payments, square card nonce's have to be encoded with the right application id--so for appropriate transactions, we need to use the Think!Chinatown id. The Square React component requires this to be passed in as a prop, s owe need to have it in the right context. This involved piping the `non_profit_location_id` down from the initial seller call to not have to make a second call for this info.

Overview:
- Seller now has `non_profit_location_id` as a property, so handling that, then passing it down through `props` down to the `SquareModal`
- Checking if Location ID's match, and if so, switching to the appropriate credential set

Supporting Details:
- Adding all Think!Chinatown credentials to the env
- Swapped the default sandbox square location ID to our sandbox app
- Ran Prettier on the files changed and also fixed some typos 😬 

Back-end changes are [in this matching PR (!167)](https://github.com/sendchinatownlove/ruby/pull/167)


.env.development
```diff
@@ -2,6 +2,7 @@
 # Be careful not the check in any secret API keys
 
 REACT_APP_API_ENDPOINT=https://staging-scl.herokuapp.com/
+# REACT_APP_API_ENDPOINT=http://localhost:5000/ # swap this for local testing
```
danthaman44, line 5: dont forget to revert before merging :)

src/components/OwnerPanel/OwnerPanel.tsx
```diff
@@ -164,6 +165,7 @@ const OwnerPanel = ({ seller }: Props) => {
         sellerId={seller.seller_id}
         sellerName={seller.name}
         costPerMeal={costPerMealDollars}
+        nonProfitLocationId={seller.non_profit_location_id}
```
danthaman44, line 168: Seems ok for now, just be careful to handle the case if this value is nil :)

danthaman44:
Hot thought: If the each seller has a `non_profit_location_id`, do we need to send `non_profit_location_id` down to the front-end at all? Can we just do a lookup on the server when the charge is being processed?

.env.development
jacoblb64: I decided to leave this in as a comment in case someone wanted it! I can remove it though if it feels unnecessary :)

jacoblb64:
> Hot thought: If the each seller has a `non_profit_location_id`, do we need to send `non_profit_location_id` down to the front-end at all? Can we just do a lookup on the server when the charge is being processed?

@danthaman44 
So that was my initial idea actually, but it felt silly to make an extra server call when we've already retrieved the full Seller information set earlier--so I thought it'd be best to pass it down and not make another web request. Also might help us save on a tid bit of server costs 😂

src/components/OwnerPanel/OwnerPanel.tsx
jacoblb64: Yup! Just tested with backend on master not having this variable and everything is fine, handling the null cases ok!

PR #209, opened 2020-07-01
Bruce/revamping voucher styling

bvillaroman:


jacoblb64:
Great refactor! This looks really great and I think is much cleaner overall. Good refactoring out of shared logic with things like the back/cancel buttons. I checked out your branch and looks like there might be some slight issues: one medium, and some others less important.

The medium one is one I saw before and tried to fix, where the MoreInfo pop-out still has it's full height even when closed, which overlays the input box and blocks tapping on it. The solution I had to this was to change the height, or maybe min-height now based on the state.
Screen Shot 2020-06-30 at 11 25 30 PM

Minor, stretched image on the landing page:
Screen Shot 2020-06-30 at 11 26 31 PM

Also minor, and maybe we can't fix if flex doesn't seem to be working the way we expect, is the footer being 'sticky' to the bottom:
Screen Shot 2020-06-30 at 11 27 38 PM

Even less important, is how things work on iPhone SE/5 sizes, which I notice because I have a tiny phone 😜, but I'm also used to things being broken for me ahha
Screen Shot 2020-06-30 at 11 26 14 PM

Anyway, we should probably get the first one fixed, the others only if you want to! Also let me know if I was testing incorrectly, just built and checked it out with the chrome inspector!

PR #212, opened 2020-07-06
added condition for empty locations

ath-chen:
- Forgot that we had a SCL donation pool merchant card. Needed to add this line since that card doesn't include any locations and was throwing errors


justintmckibben:
I'd be happy to update the model on the backend return only the first location! @danthaman44

PR #214, opened 2020-07-12
Create UI component for campaign item

danthaman44:
-Currently using placeholder static data
-Need to update progress bar to indicate new UI. Will work with @bvillaroman 
-Only works on desktop for now. Need to add responsive support for mobile. Will work with @bvillaroman 

Screen Shot 2020-07-11 at 10 26 04 PM


justintmckibben:
Some updates have been page to make it show # meals vs $

danthaman44:
@justintmckibben Ahh yeah I have that as one of the followup items. I need to chat with Bruce with whats the best way to update the progress bar. We could either modify the existing component, or fork a new component

PR #215, opened 2020-07-16
Created progress bar component for campaigns

danthaman44:
Test plan:
Screen Shot 2020-07-15 at 11 20 16 PM

Screen Shot 2020-07-15 at 11 37 15 PM

### DIFF
--- a/src/components/MerchantsPage/gam/CampaignListItem.tsx
+++ b/src/components/MerchantsPage/gam/CampaignListItem.tsx
@@ -1,68 +1,52 @@
 import React from 'react';
-import ProgressBar from '../../ProgressBar';
+import CampaignProgressBar from './CampaignProgressBar';
 import styled from 'styled-components';
 import {
-    smallScreens,
-    tabletScreens,
+  smallScreens,
+  tabletScreens,
 } from '../../../utilities/general/responsive';
 import campaignDefaultImage from '../images/campaign_default.png';
 import apexLogo from '../images/apex-logo.png';
 import melonpannaLogo from '../images/melonpanna-logo.png';
 
 // In the final implementation, campaign will be object declared in types.ts
-const CampaignListItem = (campaign: String) => {
-    return (
-        <Container>
-            <ColumnContainer>
-                <img
-                    src={campaignDefaultImage}
-                    alt="campaign_image"
-                />
-            </ColumnContainer>
-            <ColumnContainer>
-                <Location>
-                    Sunset Park, Brooklyn
-                    </Location>
-                <Name>
-                    Melonpanna Tea & Shot x APEX for the Youth
-                    </Name>
-                <Description>
-                    Partnering with APEX for the Youth, we hope to raise 200 meals for underserved Asian and immigrant youth from low-income families.
-                    </Description>
-                <TimeStamp>
-                    Last contribution made 1h ago
-                    </TimeStamp>
-                <ProgressBar
-                    amountRaised={100}
-                    targetAmount={1000}
-                    progressBarColor={'#CF6E8A'}
-                    numContributions={20}
-                    numDonations={10}
-                    numGiftCards={10}
-                    donationAmount={50}
-                    giftCardAmount={50}
-                />
-            </ColumnContainer>
-            <ColumnContainer>
-                <ImagesContainer>
-                    <img
-                        src={apexLogo}
-                        alt="merchant_logo"
-                    />
-                    <img
-                        src={melonpannaLogo}
-                        alt="distributor_logo"
-                    />
-                </ImagesContainer>
-                <Button className='button--filled'>
-                    Visit merchant
-                    </Button>
-                <Button className={'button--outlined'}>
-                    Gift a meal
-                    </Button>
-            </ColumnContainer>
-        </Container>
-    );
+interface Props {
+  campaign: String;
+}
+
+const CampaignListItem = (campaign: Props) => {
+  return (
+    <Container>
+      <ColumnContainer>
+        <img src={campaignDefaultImage} alt="campaign_image" />
+      </ColumnContainer>
+      <ColumnContainer>
+        <Location>Sunset Park, Brooklyn</Location>
+        <Name>Melonpanna Tea & Shot x APEX for the Youth</Name>
+        <Description>
+          Partnering with APEX for the Youth, we hope to raise 200 meals for
+          underserved Asian and immigrant youth from low-income families.
+        </Description>
+        {/* Testing values */}
+        <CampaignProgressBar
+          isActive={true}
+          numContributions={73}
+          targetAmount={100}
+          progressBarColor={'#CF6E8A'}
+          lastContributionTime={new Date('07/21/2020 20:05:00')}
+          endDate={new Date('07/23/2020')}
+        />
+      </ColumnContainer>
+      <ColumnContainer>
+        <ImagesContainer>
+          <img src={apexLogo} alt="merchant_logo" />
+          <img src={melonpannaLogo} alt="distributor_logo" />
+        </ImagesContainer>
+        <Button className="button--filled">Visit merchant</Button>
+        <Button className={'button--outlined'}>Gift a meal</Button>
+      </ColumnContainer>
+    </Container>
+  );
 };
 
 export default CampaignListItem;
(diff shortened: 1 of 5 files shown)


src/components/MerchantsPage/gam/CampaignProgressBar.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const progressPercent = Math.floor((numContributions / targetAmount) * 100);
+
+  const timeUntilEnd = endDate.getTime() - Date.now();
+  const daysUntilEnd = timeUntilEnd / (1000 * 3600 * 24);
+
+  return (
+    <ProgressBarContainer>
+      {hoursSinceLastContribution < 1 && (
+        <TimeStamp>
+          Last contribution made {Math.floor(minutesSinceLastConstribution)}m
```
justintmckibben, line 37: We should translate these static strings

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
```diff
@@ -0,0 +1,135 @@
...
+          <ProgressTextContainer color={progressBarColor}>
+            {progressPercent}% to target
+          </ProgressTextContainer>{' '}
+          ({numContributions} out of {targetAmount} meals)
+        </div>
+        {daysUntilEnd > 1 && (
+          <div>
+            Ends in{' '}
+            <ProgressTextContainer color={progressBarColor}>
+              {Math.ceil(daysUntilEnd)}
```
justintmckibben, line 73: What happens if it goes over but the campaign is still active?

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
```diff
@@ -0,0 +1,135 @@
...
+`;
+
+const TimeStamp = styled.div`
+  font-family: Open Sans;
+  font-style: normal;
+  font-weight: normal;
+  font-size: 13px;
+  line-height: 18px;
+  letter-spacing: 0.02em;
+  color: #9e9e9e;
```
justintmckibben, line 133: Are these colors not stored in a constant somewhere?

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
danthaman44: Screen Shot 2020-07-15 at 11 46 02 PM

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
jacoblb64: I put some colors to be shared in `src/index.scss`. I think that's a good place to be sharing these if you want to put this and/or others there.

src/components/MerchantsPage/gam/CampaignListItem.tsx
```diff
@@ -1,68 +1,51 @@
...
+  return (
+    <Container>
+      <ColumnContainer>
+        <img src={campaignDefaultImage} alt="campaign_image" />
+      </ColumnContainer>
+      <ColumnContainer>
+        <Location>Sunset Park, Brooklyn</Location>
+        <Name>Melonpanna Tea & Shot x APEX for the Youth</Name>
+        <Description>
+          Partnering with APEX for the Youth, we hope to raise 200 meals for
```
jacoblb64, line 27: nit and probably unrelated as I think this is placeholder text? But isn't it "Apex for Youth", without the "the"?

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+const CampaignProgressBar = ({
+  numContributions,
+  targetAmount,
+  progressBarColor,
+  lastContributionTime,
+  endDate,
+}: Props) => {
+  const timeSinceLastContribution = Date.now() - lastContributionTime.getTime();
+  const minutesSinceLastConstribution = timeSinceLastContribution / (1000 * 60);
```
jacoblb64, line 20: Maybe better to have these as props on one object? Also there is a [`getMinutes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMinutes) function on date, not sure if that's what you need.

Would also avoid the typo here 😂 "con**st**ribution"

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
jacoblb64: I feel like you can just have one line here instead of these three conditions
```suggestion
          <TimeStamp>
          Last contribution made {dateAsPresentable(this.lastContributionTime)} ago
        </TimeStamp>
```

where `dateAsPresentable` is some function that returns the right formatted string that's appropriate. That'll reduce the amount of duplicated code by a longshot here, and makes that date formatting reusable and easy to improve upon

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
```diff
@@ -0,0 +1,135 @@
...
+      )}
+      <TargetAmountBar className="progress-bar">
+        <CurrentProgressBar
+          style={{
+            width: `${progressWidth(numContributions, targetAmount)}%`,
+            backgroundColor: progressBarColor,
+            //defaults to default color if no color is passed in
+          }}
+        >
+          {' '}
```
jacoblb64, line 59: I've noticed you've got a few of these. Maybe something to better account for in margin/padding?

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+const TargetAmountBar = styled.div`
+  background-color: #dedede;
+  height: 12px;
+  margin-bottom: 15px;
+`;
+
+const CurrentProgressBar = styled.div`
+  background-color: #dd678a;
+  z-index: 5;
```
jacoblb64, line 117: Why this z-index here?

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
jacoblb64: Or actually just refactor this whole logic maybe to a separate function as it seems you're only using this to created a formatted string. See my other comment below.

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
jacoblb64: https://momentjs.com/ might actually be able to do just this for you

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
danthaman44: Hmm good question. Right now this will just say "Ended". Lets sync with Nanxi about this offline, this is a major edge case that warrants thought.

danthaman44: My thinking is that if we want to keep a campaign active, we can just push out the end date in the DB? Then we don't hit this edge case

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
danthaman44: Ahh yeah I copied this alot of this code over from ProgressBar. Will remove them

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
danthaman44: Same, copied this over. Removing it.

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
danthaman44: Done! Added these colors to src/index.scss. I think we would need additional eng work to export the color constants to be available in components, but this is a good start

src/components/MerchantsPage/gam/CampaignListItem.tsx
danthaman44: Ah yeah this placeholder. Will remove this once we hook this up with data from the backend.

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
danthaman44: Ok based on our Slack convo, gonna update this to display "0 days" if the campaign has reached its end date but is still active

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
danthaman44: Ok I wrote an anonymous function to handle the date presentation...thanks for the suggestion!

PR #219, opened 2020-07-25
Created updated entry point component for pool donations

danthaman44:
- Will use this as the template for the new GAM entry point box as well
- Not adding this to the Merchant page until the new GAM experience is ready

Desktop:
Screen Shot 2020-07-25 at 3 57 59 PM

Mobile
Screen Shot 2020-07-25 at 3 57 48 PM

### DIFF
--- a/src/locales/en/translation.json
+++ b/src/locales/en/translation.json
@@ -94,5 +94,10 @@
       "Website": "Website",
       "Menu": "Menu"
     }
+  },
+  "donationBox": {
+    "header": "Not sure who to donate to?",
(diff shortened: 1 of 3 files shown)


src/locales/en/translation.json
```diff
@@ -94,5 +94,10 @@
       "Website": "Website",
       "Menu": "Menu"
     }
+  },
+  "donationBox": {
```
jacoblb64, line 98: We already have a Donation Pool section here, though the text is slightly different. Maybe we can at least use the same Chinese?

PR #221, opened 2020-07-27
Abdul/queens fix

akill13:
This is a hard coded fix to ensure that flushing restaurants show under queens filtering. Will make it better later on but this is to make sure we aren't blocking people from donating to a certain restaurant


src/components/MerchantsPage/MerchantsPage.tsx
```diff
@@ -57,9 +57,15 @@ const MerchantsPage = (props: Props) => {
...
-      const result = sellers.filter(
+      let result = sellers.filter(
         (store: any) => store!.locations.length > 0 && store!.locations[0].city === type
       );
+      if (type === 'Floral Park') {
+        const flushing = sellers.filter(
+          (store: any) => store!.locations.length > 0 && store!.locations[0].city === 'Flushing'
+        );
+        result.push(...flushing);
+      }
```
justintmckibben, line 68: this is good quick fix! We should probably add a borough attribute to the backend. Let's pair on this tomorrow

PR #222, opened 2020-07-27
Merchant Voucher Dashboard

jacoblb64:
Here's a good first pass at the dashboard for merchants to see all the vouchers for their business in real time. Right now we're working with the idea that @justintmckibben and I came up with to have a unique URL that the merchants can access. He put in the back-end support for that here: https://github.com/sendchinatownlove/ruby/pull/175

I tried to follow the Figma designs as much as possible, but we're not returning quite all the data that is laid out, and we don't have any support for interactivity through the api right now. I think this is good for now, and we can revisit those features if/when Vouchers become more important for us.

I did some searching for table frameworks, and there a lot of really powerful ones with tons of customization, but [react-fiterable-table](https://github.com/ianwitherow/react-filterable-table) seemed to do pretty much everything we wanted out of the box. It hasn't been updated for about a year and has some console logs about deprecated methods, but I think that's ok for now. It expected bootstrap css, so I've added that for this page and done a bunch of custom styling on top of it. Again, we can revisit if/when vouchers become more important. For now it gives us really easy sorting, filtering, and pagination!

The page is pretty responsive at widths ~700px and wider. Unfortunately the table doesn't really work that will when narrow. It should work well on a tablet.

Points for discussion / changes:

- I made the url ~~`portal/[seller_id]/[secret]`, this could also be `dashboard/` or something else, thoughts?~~ **UPDATED** to `[seller_id]/dashboard/[secret]`
- I've set 'Last Updated' to just be the current date as we make the query live--though any other suggestions?
- I set pagination breaks at 20 items as it seemed reasonable
- Take a look at the no records found and no records returned text

Future work:
- ~~Needs chinese / translation~~ DONE
- Support new columns in backend and here
- Editability API's and front-end
- Different table framework?

Screenshots:
Screen Shot 2020-07-28 at 11 45 18 AM
Screen Shot 2020-07-28 at 11 45 43 AM


src/utilities/api/interactionManager.ts
```diff
@@ -118,6 +118,11 @@ export const updateVoucher = async (id: string, amount: number) =>
     .then((res) => res)
     .catch((err) => err);
 
+export const getMerchantGiftCards = async (seller_id: string, secret: string) =>
+  axios
+    .get(sellers + seller_id + '/gift_cards/' + secret)
+    .then((res) => res) // don't catch error, throw it up the stack
```
justintmckibben, line 124: :puke: 😂

src/components/App/App.tsx
```diff
@@ -84,6 +87,9 @@ const App = () => {
               <VoucherRedemptionPage />
             </VoucherProvider>
           </Route>
+          <Route path="/portal">
```
justintmckibben, line 90: I'd do :seller_id/dashboard/:secret

jacoblb64: sure, works for me!

src/utilities/api/interactionManager.ts
jacoblb64: 😂  I wanted to differentiate from all the other requests that have a catch block

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
```diff
@@ -0,0 +1,144 @@
+import React, { useState, useEffect } from 'react';
+import { useHistory } from 'react-router-dom';
+import {
+  getMerchantGiftCards,
+  getSeller,
+} from '../../utilities/api/interactionManager';
+import Loader from '../../components/Loader/Loader';
+import styles from './styles.module.scss';
+import ErrorPage from '../../components/404Page';
+const FilterableTable = require('react-filterable-table');
```
danthaman44, line 10: Is this something we can lazy import? Seems like a large library

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
```diff
@@ -0,0 +1,144 @@
...
+    secretId: urlParams[2],
+  };
+
+  const [error, setError] = useState(false);
+  const [loading, setLoading] = useState(true);
+  const [giftCards, setGiftCards] = useState<any | null>();
+  const [seller, setSeller] = useState<any | null>();
+
+  const renderAmount = (props) => {
+    return Intl.NumberFormat('en-US', {
```
danthaman44, line 27: Good job on the I18n!

src/utilities/api/interactionManager.ts
```diff
@@ -118,6 +118,11 @@ export const updateVoucher = async (id: string, amount: number) =>
     .then((res) => res)
     .catch((err) => err);
 
+export const getMerchantGiftCards = async (seller_id: string, secret: string) =>
+  axios
+    .get(sellers + seller_id + '/gift_cards/' + secret)
```
danthaman44, line 123: Does it make sense to do some kind of hashing to encrypt the secret before sending over the network? Lmk if you think thats over-engineering  😂

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
danthaman44: Ahh nvm I see you lazy import in App.tsx :).

src/components/App/App.tsx
```diff
@@ -34,6 +34,9 @@ const ErrorPage = lazy(() => import('../404Page'));
 const VoucherRedemptionPage = lazy(() =>
   import('../../pages/VoucherRedemption')
 );
+const MerchantGiftCardsPage = lazy(() =>
```
danthaman44, line 37: I'd rather we name this MerchantVouchersPage to keep it consistent with VoucherRedemptionPage, but up to you!

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
```diff
@@ -0,0 +1,144 @@
...
+  getSeller,
+} from '../../utilities/api/interactionManager';
+import Loader from '../../components/Loader/Loader';
+import styles from './styles.module.scss';
+import ErrorPage from '../../components/404Page';
+const FilterableTable = require('react-filterable-table');
+
+const MerchantGiftCards = () => {
+  const params = useHistory();
+  let urlParams = params.location.pathname.match(/\/[^/]+/g) as string[];
```
danthaman44, line 14: If you can find a way to combine this with the line below, that would be super slick :)

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
```diff
@@ -0,0 +1,144 @@
...
+    sellerId: urlParams[0],
+    secretId: urlParams[2],
+  };
+
+  const [error, setError] = useState(false);
+  const [loading, setLoading] = useState(true);
+  const [giftCards, setGiftCards] = useState<any | null>();
+  const [seller, setSeller] = useState<any | null>();
+
+  const renderAmount = (props) => {
```
danthaman44, line 26: consider declaring a type for props? Same for the other anonymous functions

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
jacoblb64: Oh yeah, don't want to load this all in the main app 😎

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
jacoblb64: Keeping it standard!

src/components/App/App.tsx
jacoblb64: Makes sense to me! I started with Gift Cards as that's what the endpoint is called, but then I realized we're showing 'Voucher' everywhere. I might change it to MerchantVoucherDashboard actually

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
jacoblb64: Yeah, these have an expected format from the underlying library... where I wish they were typed, but I can at least define them here
https://github.com/ianwitherow/react-filterable-table

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
```diff
@@ -0,0 +1,145 @@
...
+      currency: 'USD',
+      maximumSignificantDigits: 4,
+    }).format(props.value / 100);
+  };
+
+  const renderDate = (props) => {
+    return new Date(props.value).toISOString().substring(0, 10);
+  };
+
+  const fields = [
```
bvillaroman, line 38: seems like a static var that we could maybe move above/below the react component, it would be easier to read inner workings of the react component with the less clutter of these static values.

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
```diff
@@ -0,0 +1,145 @@
...
+      setGiftCards(giftCardData);
+      setSeller(sellerData);
+    } catch {
+      setError(true);
+    }
+    setLoading(false);
+  };
+
+  useEffect(() => {
+    setLoading(true);
```
bvillaroman, line 90: this call shouldnt be necessary if loading is already true when the component is mounted onto the DOM

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
```diff
@@ -0,0 +1,145 @@
...
+  return (
+    <>
+      {loading ? (
+        <Loader isPage={true} />
+      ) : error ? (
+        <ErrorPage menuOpen={false} />
+      ) : (
+        <>
+          <div className={styles.header}>
+            <h1>Voucher Tracker <span className={styles.noBreak}>礼品券记录</span></h1>
```
bvillaroman, line 105: im not sure how the design was for this, but maybe we should think about using i18n for these headers? But then again this is MVP so its a good fit for now.

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
```diff
@@ -0,0 +1,145 @@
...
+            <div className={styles.metadataBlock}>
+              <h1>TOTAL BALANCE <span className={styles.noBreak}>总结余</span></h1>
+              <h2>
+                {renderAmount({
+                  value: giftCards.reduce((acc, cur) => acc + cur.value, 0),
+                })}
+              </h2>
+            </div>
+          </div>
+          <FilterableTable
```
bvillaroman, line 126: maybe we can make our own Filter Table so we can use one less dependency, ngl alot of these props are giving me anxiety LOL. But this is good for now too.

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
jacoblb64: Ah good catch, I changed loading to be true by default and didn't remove this 👍

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
jacoblb64: I was initially going to use i18n, but the design actually calls for the page to be bilingual, with both on the page at the same time. I think it's spacious enough to work. The `noBreak` here was because the default word-wrapping was cutting in the middle of the chinese, where I think it makes more sense to drop it all to the second line.
https://www.figma.com/file/I0XPeX6re8t14J4nZOvVNp/Send-Chinatown-Love?node-id=4130%3A6669

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
jacoblb64: 😬  It's definitely a little anxiety-inducing. I think given that there's a discussion on how important gift cards might actually turn out to be, I thought I'd try to reduce the time to at least an MVP. I think when we want to make these fields editable (and add buttons, etc) we should either roll our own, or find a more modern table system.

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
jacoblb64: Done 👌 . It's nice and const now too

src/pages/MerchantGiftCards/MerchantGiftCards.tsx
jacoblb64: Good call! I restructured the file to be a bit more readable now I think 👌

src/utilities/api/interactionManager.ts
jacoblb64: In theory maybe we could, but really this isn't the most secure way of doing this at all. It's kind of a "security through obscurity" approach, but the stakes are pretty low here anyway. The server doesn't support anything that fancy yet, so this is the best we've got. I think if vouchers become a more important part of our organization we'll probably implement some kind of more formal authentication system.

jacoblb64:
I also removed the expiration as I realized it wasn't in the Figma design. The expiration for everything seems to be not till at least 2021 anyway, so I think it's better to remove noise and not confuse merchants.

src/utilities/api/interactionManager.ts
danthaman44: "security through obscurity" 🤣

PR #231, opened 2020-08-07
Olivia/gam pool modal

oliviasztanga:
I wanted to open a PR to get feedback.

**Things I did:**
- Corrected some styling I wasn't happy with earlier
- Moved the red GAM instructions subheading to a separate file so I could reuse it inside of the modal
- Modified the GAM modal to display the pooled donation modal if `props.sellerId` is falsy
- Added some conditional styling to the main page to render the no active campaigns box

**Points for feedback/discussion:**
- Would it be better to just create a new modal form, even if it would repeat most of the code?
- That CSS file is getting really long... what am I doing wrong?

Note: Would definitely love to talk through the pooled payment workflow.

**Screenshots:**

Page Layout - Desktop
Desktop

Page Layout - Tablet
Tablet

Page Layout - Mobile
Mobile

No Active Campaigns - Desktop
No Active Campaigns Desktop

No Active Campaigns - Tablet
No Active Campaigns Tablet

No Active Campaigns - Mobile
No Active Campaigns Mobile

Pool Modal - Desktop
Pool Modal Desktop

Pool Modal - Phone
Pool Modal Phone


src/components/MerchantsPage/gam/styles.module.scss
```diff
@@ -176,3 +183,115 @@
...
+  grid-template-columns: 2fr 1fr;
+  padding: 30px;
+
+  @media (max-width: 900px) {
+    grid-template-columns: 1fr;
+    margin-bottom: 15px;
+
+    & > div {
+      margin-bottom: 20px;
+    }
```
justintmckibben, line 200: I prefer not to have selectors like this, because then you tie yourself a rigid nesting structure, and people can accidentally break your code by adding an extra div or changing this div to a span. Instead I'd recommend just creating a separate class for this div

src/components/MerchantsPage/gam/styles.module.scss
```diff
@@ -176,3 +183,115 @@
...
+  }
+}
+
+.noCampaignText {
+  :first-child {
+    font-style: normal;
+    font-weight: 600;
+    font-size: 24px;
```
justintmckibben, line 208: To help reduce the size of css files in the future you might could refactor a lot of these font stylings into scss mixins

src/components/MerchantsPage/gam/styles.module.scss
```diff
@@ -176,3 +183,115 @@
...
+}
+
+.giftButton {
+  justify-self: flex-end;
+  text-align: center;
+  width: 240px;
+  letter-spacing: 0.15em;
+  cursor: pointer;
+  margin-bottom: 18px;
+  font-family: 'Open Sans', 'Helvetica Neue', sans-serif;
```
justintmckibben, line 234: Same here a mixin could help

justintmckibben: Or maybe something like TypeSmall TypeMedium TypeHeader, etc. like Google Docs does

src/components/MerchantsPage/gam/styles.module.scss
```diff
@@ -176,3 +183,115 @@
...
+    font-size: 11px;
+  }
+
+  :first-child {
+    text-transform: uppercase;
+  }
+
+  :last-child {
+    font-weight: 600;
+  }
```
justintmckibben, line 296: These first and last child things are starting to be a bit unclear to me since it's used a lot here. Would it be better to just define an explicit class for the first and last class of these steps?

src/components/MerchantsPage/gam/GiftAMealPage.tsx
```diff
@@ -34,43 +44,50 @@ const GiftAMealPage = (props: Props) => {
...
+            <CampaignListItem campaign={''} />
+          ))}
+        </>
+      ) : (
+        <div className={styles.noCampaign}>
+          <div className={styles.noCampaignText}>
+            <p>No active campaigns? You can still donate!</p>
+            <p>
+              Meal donations here will automatically be applied to our next
+              campaign.
```
justintmckibben, line 65: Could you extract these strings into i18n?

danthaman44:
The updated UI looks fantastic! Regarding your questions:

> Would it be better to just create a new modal form, even if it would repeat most of the code?
Let's create a new form for now. After we launch it and verify that there are no bugs, we can make a task to replace the old form with the new form. Especially since the new form looks much nicer :) 

> That CSS file is getting really long... what am I doing wrong?
Not 100% sure about this either...we can consider moving to styled components? @bvillaroman any thoughts on this?

danthaman44:
CampaignListItem now has access to the merchant data from the backend! So if you want, you can wire your new GAM flow to the "Gift A Meal" button.

src/components/MerchantsPage/gam/styles.module.scss
```diff
@@ -19,7 +19,7 @@
   grid-template-rows: 300px 215px;
 }
 
-.main {
+.header_text {
```
jacoblb64, line 22: nit: I think we've been using Pascal Case for our classes, not snake--but no need to change unless you're doing something else too

src/components/ModalBuyMeal/ModalBuyMealPool.tsx
```diff
@@ -0,0 +1,121 @@
...
+export interface Props {
+  purchaseType: string;
+  sellerId: string;
+  sellerName: string;
+  costPerMeal: number;
+}
+
+export const Modal = (props: Props) => {
+  const { t } = useTranslation();
+  const dispatch = useModalPaymentDispatch();
```
jacoblb64, line 23: Watch out, I think you and @bvillaroman will have to reconcile who merges first, as I think after his change goes in, this will have to be `useModalPaymentDispatch(null)`

src/components/MerchantsPage/gam/styles.module.scss
oliviasztanga: Yes, this is a really good point! Will fix.

src/components/MerchantsPage/gam/GiftAMealPage.tsx
oliviasztanga: Yes! I'll convert all of the text this evening as well.

src/components/MerchantsPage/gam/styles.module.scss
oliviasztanga: Noted. I can definitely take a look over and fix these.

oliviasztanga:
Thanks so much for the feedback, @danthaman44, @justintmckibben, and @jacoblb64! Sorry for delayed response -- my internet was out all weekend. This was really helpful.

I'm going to go ahead and:
- refactor CSS with attention to: (a) nonspecific selectors, (b) places I can make use of mix ins, and (c) adhering to PascalCase
- wire the GAM flow to the "Gift A Meal" button on the `CampaignListItem` component
- create separate form for the GAM flow 
- extract all text to i18n

Will definitely check in with @bvillaroman before merging!

PR #232, opened 2020-08-07
Hover on merchant progress bar/format dollar amounts

aliceluo2048:
**Changes made**:
1.	Vouchers/Donations hover when cursor is on progress bar (instead of the supporters text)
2.	Formatted the voucher and donations dollar amount to be consistent with amount raised and target amount below it (before: $2960, after: $2,960) . 

**Before**: 
image

**After**:
hoverprogressbar

Suggestions:
1.	Do we also want to round the amounts in the box (e.g. $916.75 to $916)?
2.	The tooltip placement covers the merchant description (on all merchants page) and the merchant name (on individual merchant page). Do want to change this to bottom or leave as it is? 

**Current (top placement)**:
imageimage

**Bottom placement**:
imageimage

### DIFF
--- a/src/components/ProgressBar/ProgressBar.tsx
+++ b/src/components/ProgressBar/ProgressBar.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import Tooltip from '@material-ui/core/Tooltip';
 import { withStyles, Theme } from '@material-ui/core/styles';
-
 import styled from 'styled-components';
 
 interface Props {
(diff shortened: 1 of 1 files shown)


justintmckibben:
Be sure to get @nanxiy's approval on the hover state before you merge

- [Approve?]

nanxiy:
I like still prefer the top placement! Because you want to view the hover details in conjunction with the total $ and support #, and would be caring less about the description/merchant name in that moment

aliceluo2048:
> Wow this is great!! Thanks so much for picking this up!

image
Just pushed a commit re: rounding the amounts in the tooltip display

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -69,15 +51,15 @@ const ProgressBar = ({
                         <b>{numGiftCards}</b> vouchers
                       </td>
                       <td>
-                        <b>${Math.floor(giftCardAmount) / 100}</b>
+                        <b>${Math.floor(giftCardAmount / 100).toLocaleString()}</b>
```
jacoblb64, line 54: Good call with `toLocaleString()`, we could even use [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) like we do [here](https://github.com/sendchinatownlove/sendchinatownlove.github.io/blob/aec06e27371864100f4bb0e5a93a77e9c9c3186c/src/pages/MerchantVoucherDashboard/MerchantVoucherDashboard.tsx#L49) if we want to be really fancy

PR #234, opened 2020-08-11
UI polish for GAM page

danthaman44:
-Render Seller logo_image_url instead of placeholder melonpanna logo in the Campaign list item
-Remove GAM button for inactive campaigns
-Added description text to video
-Updated with @nanxiy 's latest changes to the progress bar to reduce text overflow

Test plan
Screen Shot 2020-08-10 at 11 16 46 PM
Screen Shot 2020-08-10 at 11 17 19 PM
Screen Shot 2020-08-10 at 11 16 58 PM

### DIFF
--- a/src/components/MerchantsPage/gam/CampaignListItem.tsx
+++ b/src/components/MerchantsPage/gam/CampaignListItem.tsx
@@ -5,7 +5,6 @@ import {
   tabletScreens
 } from '../../../utilities/general/responsive';
 import campaignDefaultImage from '../images/campaign_default.png';
-import melonpannaLogo from '../images/melonpanna-logo.png';
 import { Campaign } from '../../../utilities/api/types';
 import { getDistributor, getSeller } from '../../../utilities';
 import { useEffect, useState } from 'react';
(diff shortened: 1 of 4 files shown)


src/components/MerchantsPage/gam/GiftAMealPage.tsx
```diff
@@ -87,7 +87,10 @@ const GiftAMealPage = (props: Props) => {
         <CampaignListItem campaign={campaign} />
       ))}
 
-      <VideoComponent videoId="FYlUOhzYlRM"></VideoComponent>
+      <div className={styles.videoContainer}>
+        <VideoComponent videoId="FYlUOhzYlRM"></VideoComponent>
+        <h5 className={styles.videoDescription}>Watch how we bring impact to our communties through Gift-a-Meal.</h5>
```
jacoblb64, line 92: nit: should this be translatable too?

src/components/MerchantsPage/gam/CampaignListItem.tsx
```diff
@@ -81,7 +81,9 @@ const CampaignListItem = (props: Props) => {
             window.location.href = '/' + merchant.seller_id;
           }}>Visit merchant</Button>
```
jacoblb64, line 82: where's the logic to turn this button black from red if the campaign is inactive? I might just be missing it

danthaman44: Ooops thats not a feature...hovering over the button turns it red. I just happened to be doing this when I took the screenshot 😂

src/components/MerchantsPage/gam/GiftAMealPage.tsx
danthaman44: Hmm yes thats on my TODO list for followups. Translating the GAM page will be a bit harder than other pages because some of the text (location, description, org names) come from the server, some other text is static.

src/components/MerchantsPage/gam/CampaignListItem.tsx
jacoblb64: Ahhh gotchya, that makes sense, thanks for clarifying! 😂

src/components/MerchantsPage/gam/GiftAMealPage.tsx
jacoblb64: Makes sense, sounds good!

PR #235, opened 2020-08-12
FIX: Donate nav link

aliceluo2048:
I noticed the bottom border under Donate in the nav bar is not showing when it's the active page. I could have changed  '/merchants' to  '/all'' in `isMerchantsPathActive` but I did some refactoring (heh). Let me know if you want me to revert to what we had. I also added some styling to make it closer to the Weebly underline.

**Currently:**
https://merchant.sendchinatownlove.com/all
image

**Now:**
image

**Compared to Weebly's underline**
image


justintmckibben:
:pray:

PR #242, opened 2020-08-18
passport vouchers to print

ath-chen:
this is what it'll look like in print view!

Screen Shot 2020-08-18 at 8 54 27 AM


ath-chen:
@jacoblb64 tried it & updated the photo, if they make it no margins & 68 scale it should fit. i just have to make a small edit to the size of the image so there's no spillage

jacoblb64:
> @jacoblb64 tried it & updated the photo, if they make it no margins & 68 scale it should fit. i just have to make a small edit to the size of the image so there's no spillage

@ath-chen, nice! I think having to fiddle with the settings is fine as it'll be us in control of printing everything out. That sounds like a good solution to me!

PR #245, opened 2020-08-20
Athena/passport

ath-chen:
Updated PR with new changes: 

- Existing user can update ig handle 
- only allow user to "view tickets" if they have one or more verified ticket 
- updated links / icons / logos / general formatting


src/pages/PassportRedemption/RedemptionSelectScreen.tsx
```diff
@@ -107,6 +107,7 @@ const PassportSelected = ({ setCurrentScreenView }: Props) => {
       >
         {allSponsors.length > 0 &&
           allSponsors.map((sponsor: any) => {
+            console.log(sponsor.logo_url);
```
justintmckibben, line 110: delete

src/pages/PassportRedemption/RedemptionSelectScreen.tsx
```diff
@@ -125,7 +126,14 @@ const PassportSelected = ({ setCurrentScreenView }: Props) => {
                 />
 
                 <SingleRewardInfo>
-                  <Text className="header">{sponsor.reward}</Text>
+                  <Text
+                    className="header"
+                    style={{
+                      fontSize: sponsor.reward.length > 20 ? '10px' : '12px',
```
justintmckibben, line 132: idk if character length is the right thing to be switching on here since each character has an inherently different width. Overflow should be the real determiner here. However, I recently made a backend change that helps solve this. Otherwise, I think it's better if each card has consistent style. cc @nanxiy @oliviasztanga 

Backend PR: https://github.com/sendchinatownlove/ruby/pull/223

src/pages/PassportRedemption/RedemptionSelectScreen.tsx
aliceluo2048, line 110:
```suggestion
```

src/pages/PassportRedemption/RedemptionSelectScreen.tsx
ath-chen: okies, let me just remove that file from my commits and we can figure that one out later tonight or smth

PR #246, opened 2020-08-20
Make GAM button full width if the merchant doesnt sell vouchers

danthaman44:
Merchant sells vouchers:
Screen Shot 2020-08-20 at 12 07 05 AM

Merchant doesn't sell vouchers:
Screen Shot 2020-08-20 at 12 08 18 AM


justintmckibben:
Good catch

PR #257, opened 2020-08-24
Update Owner panel to use Campaigns API

danthaman44:
Migrate the owner panel from using Seller model data to Campaign model data for GAM

Test plan:
Verified GAM flow works from Owner Panel
Verified charge request is send to backend with correct campaign_id

Screen Shot 2020-08-23 at 11 34 43 PM


justintmckibben:
YAYYY

PR #265, opened 2020-08-27
update track screen with insta handles and formatting

ath-chen:
* DELETED LAST PR - made a new one that's cleaner and removed console.logs & changes to header styling

- update insta handle for existing users 
- cannot view tickets unless there is min one ticket 
- general formatting and updates to links / logos / icons 

Screen Shot 2020-08-27 at 4 52 46 PM


src/pages/PassportRedemption/TrackScreen.tsx
```diff
@@ -156,40 +168,42 @@ const Track = ({ setCurrentScreenView }: Props) => {
...
+                Instagram Handle (for Digital Giveaway)
+              </Label>
+              <SupporterTooltip
+                title={
+                  <React.Fragment>
+                    <ToolTipTable>
+                      <tbody>
+                        <tr>
+                          To be entered into our weekly Digital Giveaways, visit 3
+                          merchants and post and tag <b>@sendchinatownlove</b>
```
aliceluo2048, line 183: When I was doing the FAQ markup, I read that `<strong>` or CSS font-weight property is preferred over `<b>` for styling text??

src/pages/PassportRedemption/TrackScreen.tsx
```diff
@@ -156,40 +168,42 @@ const Track = ({ setCurrentScreenView }: Props) => {
...
+                  <img src={CrawlInfoIcon} alt="instagram-crawl-info" />
+                </div>
+              </SupporterTooltip>
+            </Row>
+            <InputField
+              name="instagram-handle"
+              type="text"
+              onChange={(e) => setinstagramHandle(e.target.value)}
+              value={instagramHandle}
+              placeholder="@"
```
jacoblb64, line 204: Are expecting users to type the `@` in themselves? I'm wondering if we should take the `$` [approach we did on the voucher page](https://github.com/sendchinatownlove/sendchinatownlove.github.io/blob/94478b8948af526e50f877a0061d55c0a7dd4ead/src/pages/VoucherRedemption/Amount.tsx#L149) and show it, but not have it be part of the value, or at restrict input to a-z only.

Just thinking about our data consistency, @justintmckibben, thoughts?

PR #268, opened 2020-08-28
Bruce/passport rebase5

bvillaroman:


src/pages/PassportRedemption/Passport.tsx
```diff
@@ -79,36 +79,40 @@ const Passport = (props: Props) => {
...
+      let tempTickets = [...tickets]; 
+      const sortedTickets = tempTickets
+        .sort((a, b) => {
+          return b.sponsor_seller_id - a.sponsor_seller_id;
+        })
+        .sort((a, b) => {
+          const dateA = a.redeemed_at ? Date.parse(a.redeemed_at) : 0;
+          const dateB = b.redeemed_at ? Date.parse(b.redeemed_at) : 0;
+          return dateB - dateA;
+        });
```
aliceluo2048, line 96: I think this would be a safer and more concise sorting algo:
```
.sort((a, b) => {
        const dateA = a.redeemed_at ? Date.parse(a.redeemed_at) : 0;
        const dateB = b.redeemed_at ? Date.parse(b.redeemed_at) : 0;
        if (dateA != dateB) {
          return dateB - dateA;
        }
        return b.sponsor_seller_id - a.sponsor_seller_id;
      })
```

PR #270, opened 2020-08-28
styling updates to trackScreen

ath-chen:
**note: i started this pr yesterday, so there are a couple changes in here. going forward i can make one change per commit 

changes in this pr:

1. all ticket codes will be sent to the backend as capitalized values 
2. bolded "for digital giveaway"
3. added "optional" under insta handle text field
4. will only be able to see the button "View My Tickets" when an email is entered 
5. updated tooltip to disappear after 30 seconds (since last touch) from 1.5 seconds lol <-- lmk if we want to make this even longer
6. changed e.key to e.keyCode !== 8; apparently androids read keyCode & not key? Not sure, need to merge and test on an android phone. but this change works on chrome so far

### DIFF
--- a/src/pages/PassportRedemption/TrackScreen.tsx
+++ b/src/pages/PassportRedemption/TrackScreen.tsx
@@ -31,7 +31,6 @@ const Track = ({ setCurrentScreenView }: Props) => {
   const [isTicketValid, setIsTicketValid] = useState(true);
 
   const [instagramHandle, setinstagramHandle] = useState('');
-  // const [contactId, setContactId] = useState('');
 
   const findOrCreateUser = async (email, viewTickets) => {
     const { data } = await getPassportEmailId(email);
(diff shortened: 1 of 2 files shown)


nanxiy:
for the tool tip, if you click outside of it will you be able to close it? I think 30 sec is good

ath-chen:
> for the tool tip, if you click outside of it will you be able to close it? I think 30 sec is good

I'm not sure. Have to try it on mobile since the experience is diff on my laptop

src/pages/PassportRedemption/TrackScreen.tsx
```diff
@@ -153,7 +155,7 @@ const Track = ({ setCurrentScreenView }: Props) => {
                 setIsTicketValid(true);
               }}
               onKeyUp={(e) => {
-                if (e.key !== 'Backspace') {
+                if (e.keyCode !== 8) {
```
justintmckibben, line 158: this is right

PR #272, opened 2020-08-29
removing autopopulated hyphen from ticket field

ath-chen:
- taking out hyphen auto population until i figure out how to make it work on androids :(

### DIFF
--- a/src/pages/PassportRedemption/TrackScreen.tsx
+++ b/src/pages/PassportRedemption/TrackScreen.tsx
@@ -80,11 +80,11 @@ const Track = ({ setCurrentScreenView }: Props) => {
     }
   };
 
-  const formatTicketCode = (code) => {
-    if (code.length === 3) {
-      setTicketCode(code + '-');
-    }
-  };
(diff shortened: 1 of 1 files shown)


src/pages/PassportRedemption/TrackScreen.tsx
```diff
@@ -80,11 +80,11 @@ const Track = ({ setCurrentScreenView }: Props) => {
...
-  const formatTicketCode = (code) => {
-    if (code.length === 3) {
-      setTicketCode(code + '-');
-    }
-  };
+  // const formatTicketCode = (code) => {
+  //   if (code.length === 3) {
+  //     setTicketCode(code + '-');
+  //   }
+  // };
```
justintmckibben, line 87: I think what you wanna do is:
1. If the length is > 3
2. Take the whole string
3. Trim off any non alphanumerics
4. Uppercase
5. And insert the dash at the third index

justintmckibben: Then all you need to do is run this at every onKeyUp

PR #273, opened 2020-08-29
olivia/passport redemption fixes 2

oliviasztanga:
I batched up a couple of smaller bug fixes here, but they each have their own commit.

**Summary of Fixes:**
- the `Claim Screen` now references the `logo_url`, not a placeholder
- I fixed the REEDEM NOW typo on the `Select Screen` screen
- <s>I removed the `reward_detail` from the `Claim Screen` and moved it to the `Select Screen`</s>
- I added `reward_detail` to the `Select Screen`
- I hid radio buttons on the `Select Screen` when the user has zero rewards available
- I enabled deselecting rewards by clicking again on a select reward, but I will check in with Nanxi to see if she wants to implement this differently

### DIFF
--- a/src/pages/PassportRedemption/RedemptionClaimScreen.tsx
+++ b/src/pages/PassportRedemption/RedemptionClaimScreen.tsx
@@ -61,7 +61,7 @@ const PassportRedemptionClaim = ({ setCurrentScreenView }: Props) => {
           return { id: ticket.id, sponsor_seller_id };
         });
       const { status } = await redeemReward(id, access_token, ticketsToRedeem);
-      // add some kind of error handling to redirect user here
+      // figure out how to handle invalid redemption with this page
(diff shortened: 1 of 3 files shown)


src/pages/PassportRedemption/RedemptionClaimScreen.tsx
```diff
@@ -34,8 +34,7 @@ const PassportRedemptionClaim = ({ setCurrentScreenView }: Props) => {
       zip_code: '',
     },
     logo_url: '',
-    reward: '',
-    reward_detail: '',
```
justintmckibben, line 38: why delete?

oliviasztanga: Ooh sorry! I think I misunderstood the wireframe. I'll add it back!

PR #284, opened 2020-08-31
Final quick fixes (hopefully)

bvillaroman:
Quick fixes solved: 

-fix background height for smaller screens (when there is scrolling) → becomes white on the bottom (Athena)

-New functionality: (two options) for states of “Instagram handle added and not added

-added extra copy to the popup about expiration time

-On iOS/iphone 11 button is sort of squished 

-On iOS/iphone 11 pop up is hidden sort of? And then scrolls under the card header (height of window was squished)

- stamps added from the beginning rather than to the end…. You can’t see the new ones you add. The new stamps should be added to the end. The new stamps are also populating under the unredeemed “Send to Email” row when they should be populating in the row below it

-make the haze more distinct so it’s more obvious that the cards are a separate scroll-y thing?

-The grey overlay should also go on top of the stamps so the whole thing is greyed out but low priority (desktop chrome)

-On both mobile (iPhone 11 ios 13.6 chrome) and desktop (chrome) the background only repeats once. (Athena)


src/pages/PassportRedemption/style.tsx
```diff
@@ -12,23 +12,24 @@ type PassportProps = {
...
-  box-shadow: 0 0 10px rgba(0,0,0,0.25);
+
+  box-shadow: 0 0 10px rgba(0, 0, 0, 0.25);
   border: 1px solid #dedede;
   border-radius: 20px;
   box-sizing: border-box;
   overflow: hidden;
 
   position: absolute;
+  // opacity: ${(props: PassportProps) => (props.mainView ? '1' : '0.85')};
```
justintmckibben, line 30: meant to comment out?

src/pages/PassportRedemption/RedemptionClaimScreen.tsx
```diff
@@ -132,7 +135,10 @@ const PassportRedemptionClaim = ({ setCurrentScreenView }: Props) => {
           <ContentContainer>
             <CardText size="15px" color="#a8192e">
               Offer must be used in:
-              <span style={{fontWeight:'bold'}}> {formatTime(timeLeft)}</span>
+              <span style={{ fontWeight: 'bold' }}>
```
justintmckibben, line 138: You can also just wrap this in <strong></strong> or give it a separate class

justintmckibben: If you want the inline block behavior you can still wrap it in a span

src/pages/PassportRedemption/style.tsx
bvillaroman: yeah for now, if we change the opacity, the background bleeds through the card, so for  now ill comment it out, but i think we can incorporate it again in the future

src/pages/PassportRedemption/RedemptionClaimScreen.tsx
bvillaroman: ill add this rq

PR #306, opened 2020-09-07
Add query pagination to passport tickets endpoint

justintmckibben:


src/pages/PassportRedemption/PassportVoucher.tsx
```diff
@@ -6,13 +6,31 @@ import {
...
+import { useLocation } from 'react-router-dom';
 import { LoaderFillerContainer } from '../../components/Loader';
 import VoucherImage from './VoucherFront.png';
 import BackImage from './VoucherBack.png';
 
 const Voucher = () => {
   const { id, tickets_secret } = useParams();
+  let params = new URLSearchParams(useLocation().search);
+
+  const convertToInt = (input: string | null) => {
```
jacoblb64, line 18: Why not `Integer.Parse()`? Is this doing something different here?

jacoblb64 (approved):
Looks good! Synced offline about sending `null`, and as long as we figure that out we're good to go!

src/pages/PassportRedemption/PassportVoucher.tsx
justintmckibben: Yeah this handles nulls where parseInt doesn't handle nulls

PR #307, opened 2020-09-08
Update payment method UX content

chelchan:


jacoblb64 (approved):
Looks good! FYI I was a bit worried about the longer header text and wanted to make sure we handled the overflow correctly, so I checked out your branch and verified. This should be good to go!

image

PR #308, opened 2020-09-08
Update payment method UX content

chelchan:


jacoblb64 (approved):
Looks good! Confirmed in your other MR that the longer text shouldn't be a problem 👍

PR #311, opened 2020-09-11
Add Think!Chinatown app id and location

justintmckibben:


jacoblb64 (approved):
Double checked that these are right!

PR #313, opened 2020-09-11
Transaction fee opt in

Kenworthe:
https://trello.com/c/BjLfqgG9/549-make-customer-pay-for-fees

- Adds checkbox to ModalAmount for customers to opt-in to covering transaction fee. Bool is: `covered_by_customer`
- Refactored ModalAmount a bit

image

image

## Questions: 
- As is, this branch will just increase the cost of the existing line item by 3% and pass that new amount to Square. Should we instead add a separate line item representing the Fee?

- ~How precisely do we want to cover the fee? (Actual formula is like `2.9% + 0.30c` according to another card). But it doesn't display nicely compared to flat 3%, since it causes amounts to be like: 0.57, 0.99, 1.68, etc.~

## Todo:
- [ ] add Chinese translations
- [ ] specs


Kenworthe:
@justintmckibben @Jeremy-Xue I'll get the build working; it was late and I just wanted to get the PR open for review first 😅

src/components/ModalPayment/SquareModal.tsx
```diff
@@ -352,10 +353,6 @@ const CheckboxContainer = styled.label`
   :hover {
     text-decoration: underline;
   }
-
-  > span {
-    padding: 9px 9px 9px 0px;
-  }
```
Kenworthe, line 358: Removing this fixes the misshapen checkbox effect on prod: 
image

src/utilities/hooks/ModalPaymentContext/constants.ts
```diff
@@ -6,19 +6,27 @@
...
+const SET_COVERED_BY_CUSTOMER = 'SET_COVERED_BY_CUSTOMER';
 const CLOSE_MODAL = 'CLOSE_MODAL';
 const CLEAR_FORMS = 'CLEAR_FORMS';
 
 const SET_SELLER_DATA = 'SET_SELLER_DATA';
 const UPDATE_SELLER_DATA = 'UPDATE_SELLER_DATA';
 
 const EMAIL_REGEX = /[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,}$/;
+
+const TRANSACTION_FEE_PERCENT = 0.03;
```
justintmckibben, line 19: The actual fee is 2.9% + $.30. We should account for the exact amount, rounded up to the nearest cent. cc @nanxiy would you prefer cent or dollar?

src/locales/en/translation.json
```diff
@@ -1,13 +1,26 @@
...
+      "donation": {
+        "header": "Donation for",
+        "body": "Your donation is secure, encrypted, and directly goes to the restaurant owner for sustaining their business."
+      },
+      "voucher": {
+        "header": "Voucher for",
+        "body": "Your payment is secure, encrypted, and directly goes to the restaurant owner for sustaining their business."
+      },
+      "cover": {
+        "body1": "When you donate via credit card, there are usually payment processing fees of 2-3% on the transaction. If you make a $100 donation, the merchant only gets $97 of it.",
```
justintmckibben, line 13: We should make this copy more succinct. I think we can remove the 2-3% and just say 3%. I think 3% is easy to wrap your mind around, even though it's actually 2.9% + $.30. cc @nanxiy

justintmckibben: I'd also remove "usually", since there is always a processing fee

src/utilities/hooks/ModalPaymentContext/types.ts
```diff
@@ -14,13 +15,15 @@ import { BrowsePageSeller } from '../../api/types';
...
   amount: string;
   customInput: boolean;
+  coveredByCustomer: boolean;
   modalView: number;
   sellerData: BrowsePageSeller;
 };
 
 export const defaultState: ModalPaymentState = {
-  amount: '5',
+  amount: '25',
```
justintmckibben, line 24: Why did we update this?

src/utilities/hooks/ModalPaymentContext/constants.ts
justintmckibben: btw in the future we'll add a Fees endpoint that will return the fees that are associated with the payment cc @Jeremy-Xue

justintmckibben:
Looking good so far!

src/utilities/hooks/ModalPaymentContext/constants.ts
nanxiy: I think rounded is fine, 3%

aliceluo2048: From Square: 
You can send an unlimited number of invoices for a fee of 2.9% + 30¢. Fees are applied at the time an invoice is paid by your customer with their credit or debit card, Apple Pay, or Google Pay.

We should get rid of credit card in the sentence since all transactions incur a fee of 2.9% + 30 cents.

src/locales/en/translation.json
aliceluo2048: FYI the merchant donate page currently has "100% of all proceeds go to the businesses, and we cover the credit card processing fees." - which contradicts what we are proposing

justintmckibben: @nanxiy can we remove the "and we cover the credit card processing fees" part? 100% of proceeds still do go to the merchant

src/utilities/hooks/ModalPaymentContext/constants.ts
justintmckibben: @nanxiy rounded to cent or dollar?

src/locales/en/translation.json
nanxiy: yeah let's remove it - Chinese version will need to be updated too, will let Chel know

src/utilities/hooks/ModalPaymentContext/constants.ts
nanxiy: round to cent - I like the transparency

src/utilities/hooks/ModalPaymentContext/types.ts
Kenworthe: @justintmckibben Whoops, I was playing around with defaulting the selected amount and forgot to change it back. I'll revert this on next push.

src/locales/en/translation.json
Kenworthe: @nanxiy Hmm. So what should the final copy be? 

`cover.body1`
- remove "usually"
- just write "3%" instead of "2-3%"
- remove "via credit card"

`contributionBar.footer`
- Remove "and we cover the credit card processing fees"

Hmm. Also, would this copy appear a bit contradictory if we state "3%", but when the $100 option is selected, the customer sees transaction fees are "$3.20"?  Likewise with the lower amounts.

nanxiy: "When you donate, there are payment processing fees of about 3% on the transaction. If you make a $100 donation, the merchant only gets $97 of it."

^ Added "about" to the copy. We can either say 'about' and do cents, or just do 3% and then round to the dollar. I think I prefer the former

Kenworthe: Agreed, I like the former better too. I'll make the changes, thanks! 👍

src/utilities/hooks/ModalPaymentContext/constants.ts
justintmckibben: rounded up btw

src/locales/en/translation.json
justintmckibben: @nanxiy @Kenworthe After looking at this PR and seeing the confusion in copy, I think we should do one of two things:
1. Always apply the 3% fee on top of the donation amount. The average donation is $5-$10 right now, which is at most a 60 cent fee, which to the donor is negligible. So if they select $10, they'll be charged $10.60.
2. Always apply the 3% fee and take it away from the donation amount. In this case, they'll be still charged $10, but in a breakdown somewhere we'll show them that the X will go to the merchant, and Y will go towards credit card processing fees.

Ultimately we want donations to be as fast and easy as possible, and reducing the amount of mental effort that needs to be taken when making a donation is important — while still keeping things transparent. At the end of the day, option 1 and 2 are the same thing, it's just a matter of perspective. 

@nanxiy would you be alright with revisiting the design of this a tad and @Kenworthe sorry for the thrash on the PR. We should have caught this in Figma before it got to this stage

Kenworthe: @justintmckibben okay as discussed earlier, I just made the changes to the precise transaction fee formula, rounded up to the nearest cent. left copy alone for now - we can do that as a fast follow.

nanxiy: @justintmckibben Can you clarity revisiting which part of the design ? did we decided if we were going with 1 or 2? I think if we go with 1 then the design stands, which is what Ken is working on right?

jacoblb64 (commented):
Looks great to me so far! Left some small comments--but overall nice refactoring and keeping this clean.

Sounds like we have some other implications to work out about rounding, etc. so holding off on the approval for now.

src/components/ModalAmount/ModalAmount.tsx
```diff
@@ -17,20 +19,56 @@ export interface Props {
...
 }
 
+const transactionFee = (amount: string) => {
+  const raw = (Number(amount) * TRANSACTION_FEE_RATE) + TRANSACTION_FEE_FLAT;
+  const roundedUp = Math.ceil(raw * 100) / 100;
+  return roundedUp.toFixed(2);
+};
+
+const formatCurrency = (value: string) => {
+  return `$${ Number(value).toFixed(2) }`;
```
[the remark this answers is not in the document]
Kenworthe: Ah, good callout! I'll consolidate into a helper or something.

src/pages/MerchantVoucherDashboard/MerchantVoucherDashboard.tsx
```diff
@@ -46,11 +47,7 @@ const MerchantVoucherDashboard = () => {
   }, []);
 
   const renderAmount = (props: FTRenderProps) => {
-    return Intl.NumberFormat('en-US', {
-      style: 'currency',
-      currency: 'USD',
-      maximumSignificantDigits: 4,
-    }).format(props.value / 100);
+    return currencyFormatter(Number(props.value), 0)
   };
```
Kenworthe, line 51: @jacoblb64 I believe the: 
```javascript
minimumFractionDigits: 0,
maximumFractionDigits: 0
```
below should accomplish the same intent of rounding to the nearest whole dollar, so the logic here should be preserved.

jacoblb64 (commented):
Looks all good from my side! Deferring to @justintmckibben and @nanxiy for final sign off on this one 👍

src/locales/en/translation.json
justintmckibben: I think that we need to revisit design here. The mocks are confusing, and are basically making the customer choose between two options that essentially do the same thing. I think we should remove the copy and the ability to choose, and instead just give the customer a breakdown of where the money is going upon checkout. Fundamentally, giving the customer this option doesn't really make any difference. 

We should either: 
1. Always charge the customer the fee, and tell them that they're being charged (like we do taxes upon checkout)
2. Always take the fee out of the amount the customer has selected to donate (which will be surfaced in a breakdown similar to taxes)

justintmckibben (approved):
I don't agree with the design here, but the changes do match the mock. I think that we should revisit design here though. cc @nanxiy @Kenworthe

src/locales/en/translation.json
nanxiy: https://www.figma.com/file/I0XPeX6re8t14J4nZOvVNp/Send-Chinatown-Love?node-id=104%3A844

Kenworthe:
Closing this PR in favor of a new branch which follows updated mocks (always charge transaction fees).

PR #314, opened 2020-09-11
Updated keys for the CREATE account

jacoblb64:
- Turns out there's a different Think!Chinatown square account that they use
- Already switched the keys in the backend heroku dashboard


justintmckibben (approved):
Thanks for doing this! Did we update Lanzhou Ramen yet?

jacoblb64:
@justintmckibben I'll update Lanzhou with the proper `non_profit_location_id` right now before merging

PR #320, opened 2020-09-15
Created UI components for Lyft Rewards

danthaman44:
Screen Shot 2020-09-14 at 11 55 35 PM
Screen Shot 2020-09-15 at 12 12 30 AM
Screen Shot 2020-09-15 at 12 12 55 AM


src/pages/PassportRedemption/LyftPromo.tsx
```diff
@@ -0,0 +1,121 @@
...
+  width: 340px;
+  margin: 0 auto;
+  height: ${(props) => props.height}px;
+  z-index: 20;
+  top: 70px;
+  display: flex;
+  flex-direction: column;
+  justify-content: space-around;
+  align-items: center;
+
```
aliceluo2048, line 76:
```suggestion
border-radius: 13px;
```

Dannnn do you hate border-radius or something. Might as well fill this empty line.

src/index.scss
```diff
@@ -106,6 +106,18 @@ h3 {
     background-color: $standard-red;
     border-color: $standard-red;
   }
+
+  &--red-outlined {
```
justintmckibben, line 110: nit: i much prefer BEM classes over modifier classes like this, but we have no established style-guide yet so this is fine

src/pages/PassportRedemption/LyftPromo.tsx
```diff
@@ -0,0 +1,121 @@
...
+    isSuccess: boolean;
+}
+
+export const LyftRewardPromo = () => {
+    const { t } = useTranslation();
+    return (
+        <LyftPromptContainer height={340}>
+            <LyftIcon src={BikeImg} />
+            <TitleRow>
+                <SubTitle bold="700">
```
justintmckibben, line 17: could we give this a classname and bold it in the CSS?

justintmckibben (approved):

src/pages/PassportRedemption/LyftPromo.tsx
```diff
@@ -0,0 +1,121 @@
...
+  backdrop-filter: blur(4px);
+`;
+
+const LyftIcon = styled.img`
+  margin-top: 10px;
+  width: 160px;
+  height: 80px;
+`;
+
+const LyftYesButton = styled(Button)`
```
aliceluo2048, line 88: The Yes, No and Close Button css are exactly the same

src/locales/en/translation.json
```diff
@@ -145,5 +145,17 @@
     "prompt": "Please enter how many meals to gift",
     "totalLabel": "Total: ",
     "next": "Next"
+  },
+  "lyftReward": {
```
jacoblb64, line 149: I think you have to add this same (english) text the cn file--otherwise it'll come up as empty strings in the zh-cn UI

jacoblb64 (changes requested):
Looks good to me! I do think you have to update the `cn/translation.json` file though, otherwise this'll be broken if someone selects that language option

src/index.scss
danthaman44: This is my first time learning of BEM...interesting :)

src/pages/PassportRedemption/LyftPromo.tsx
danthaman44: Done

src/locales/en/translation.json
danthaman44: ooh yeah good catch, updated the chinese!

src/pages/PassportRedemption/LyftPromo.tsx
danthaman44: Not true! They have diff margins 😂

src/pages/PassportRedemption/LyftPromo.tsx
danthaman44: Done

src/pages/PassportRedemption/LyftPromo.tsx
jacoblb64: Can we refactor the shared styles into one class then? 😂  Not a blocker, but I do find this an issue with using `styled` like this often.

jacoblb64 (approved):
Thanks for adding the cn text even if it's just in English! Left one nit 😉

PR #322, opened 2020-09-16
Refactor Lyft css

aliceluo2048:
Remove duplicate css that is based on the base-button from index.scss as well as same css in the lyft buttons
Min-width - comes from the index.scss which is why the buttons are long
Take it or leave these changes
claim
rewards


danthaman44 (approved):
Looks clean!

PR #328, opened 2020-09-21
Using react-markdown to render the merchant summaries and stories

jacoblb64:
Background
- [react-markdown](https://github.com/rexxars/react-markdown) seems to be the cleanest, easiest way to achieve this
- It also escapes HTML and keeps us safe from XSS attacks
- It seems a little rusty in terms of maintenance, but I still think it's worth adopting
- All of our text is backwards-compatible, so nothing should break

Work completed
- Used `react-markdown` to render the story and summary sections of the merchant page
- Added some standard brand styling for links, everything else is standard
- Shared styling as best as we could

Next steps
- Update the stories in our database with markdown

Screenshots:
markdown

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "react-filterable-table": "^0.3.434",
     "react-ga": "^2.7.0",
     "react-i18next": "^11.7.0",
+    "react-markdown": "^4.3.1",
     "react-number-format": "^4.4.1",
     "react-redux": "^7.2.1",
     "react-router-dom": "^5.2.0",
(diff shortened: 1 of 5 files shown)


justintmckibben (approved):
nice super easy win

jacoblb64:
Was looking at our prod site and realized we also have text in the summary, especially with the annoyingly unclickable link for 47 division street. Added the same component there and shared the styling.

@justintmckibben take a look again!

aliceluo2048 (approved):

PR #332, opened 2020-09-26
Wilsonj806/refactor giveaway popover

wilsonj806:
- Removed redundant styling
- Refactored how the IG button was rendered so it doesn't do negative pixels
- Converted pixel values into even/ integer values whenever possible
  - the PopoverContent decimal value is for centering it
  - the IG toggle button positioning probably can't be moved around too much since it also has to line up with the button inside the popover
- haven't dealt with localization yet, need to ask about it

### DIFF
--- a/src/pages/PassportRedemption/GiveawayPopover.tsx
+++ b/src/pages/PassportRedemption/GiveawayPopover.tsx
@@ -17,9 +17,6 @@ interface TextProps {
   icon ?: boolean
 }
 
-interface IgButtonProps {
-  disabled ?: boolean
-}
 interface GiveawayData {
   weekly_giveaway_entries: String
   number_of_visits_left : undefined | number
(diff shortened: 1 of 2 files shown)


src/pages/PassportRedemption/GiveawayPopover.tsx
```diff
@@ -1,121 +1,131 @@
...
+  async function setGiveawayCount(contactId: string) {
+    const { data: allSellers } = await getAllParticipatingSellers();
+    const {
+      data: { weekly_giveaway_entries, unique_seller_tickets },
+    } = await getGiveawayTicketsForContact(contactId);
+    setData({
+      weekly_giveaway_entries,
+      number_of_visits_left: allSellers.length - unique_seller_tickets,
+    });
+  }
```
aliceluo2048, line 39: Did we end the component early?

wilsonj806: Yeah @justintmckibben told me I could merge it in to get it out, and then submit another PR with refactors

justintmckibben (approved):

PR #335, opened 2020-10-03
Create Page enum for pages (all, seller, merchant, etc.)

chrisng93:
Create Page enum for pages (all, seller, etc.) rather than using strings ('all', 'seller', etc.) for the pages. Also did a bit of alphabetizing and rearranging imports on the files I touched.

### DIFF
--- a/src/consts.ts
+++ b/src/consts.ts
@@ -1,3 +1,11 @@
+enum Page {
+  All,
+  Error,
+  GiftAMeal,
+  Merchants,
+  Seller,
+};
+
 const SquareErrors = {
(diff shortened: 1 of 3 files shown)


src/consts.ts
```diff
@@ -1,3 +1,11 @@
+enum Page {
```
chrisng93, line 1: i put this here since it seemed like a reasonable shared file to put it in, but please let me know if there's a better place! i'm new to the codebase

justintmckibben (approved):
Looks great!

PR #337, opened 2020-10-04
Share social media link hardcodes across `Footer` and `Track`

chrisng93:
Share social media link hardcodes across `Footer` and `Track`. I'm new to the codebase and have been taking some time to look through the code. I noticed that we're using the same social media links across the footer and track pages. I figured we could probably reuse some code here

### DIFF
--- a/src/pages/PassportRedemption/TrackScreen.tsx
+++ b/src/pages/PassportRedemption/TrackScreen.tsx
@@ -1,11 +1,11 @@
-import React, { useState } from 'react';
-import styled from 'styled-components';
 import Tooltip from '@material-ui/core/Tooltip';
 import { withStyles, Theme } from '@material-ui/core/styles';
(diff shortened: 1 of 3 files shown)


src/pages/PassportRedemption/TrackScreen.tsx
```diff
@@ -97,19 +98,6 @@ const Track = ({ setCurrentScreenView }: Props) => {
     },
   }))(Tooltip);
 
-  const socialMediaLinks = [
```
chrisng93, line 100: just curious if there's a reason why we don't include the email in these social media links but we include them in the links for the footer

justintmckibben: It wasn't on the designs cc @nanxiy

justintmckibben (approved):

PR #338, opened 2020-10-04
Create `MAILTO_URL` const and use it across the app

chrisng93:
Create `MAILTO_URL` const and use it across the app. I noticed we were hardcoding the mailto address in the app, so I created a const for it instead and used that in the app.


justintmckibben (approved):
Thanks for cleaning up!

PR #339, opened 2020-10-06
Voucher redemption flow bug fixes

chrisng93:
Fixes a couple bugs we were seeing in the voucher flow:
- Voucher redemption landing page - fix avatar url, fallback to SCL's logo if merchant doesn't have a logo
- Voucher confirmation page - fix "(?)" component styling for single use vouchers, fix avatar for non-single use vouchers

Confirmation page - component styling for non-single use vouchers (this was never broken - just using this as a comparison for below)
Screen Shot 2020-10-05 at 4 18 21 PM

Confirmation page - fixed component styling for single use vouchers
Screen Shot 2020-10-05 at 4 16 02 PM

Redemption page - fallback to SCL logo for merchants w/o a logo
Screen Shot 2020-10-05 at 4 49 40 PM

Redemption page - fix merchant avatar
Screen Shot 2020-10-05 at 4 58 43 PM

Ticket: https://trello.com/c/c65rvoDp/604-fix-broken-links-and-other-ui-bugs-in-gift-card-redemption-flow


src/pages/VoucherRedemption/StoreBanner.tsx
```diff
@@ -66,6 +59,7 @@ const Header = styled.div`
   line-height: 30px;
 `;
 const OwnerImage = styled.img`
+  object-fit: cover;
```
chrisng93, line 62: Added this because the merchant logo was getting squished. Looks like a lot of our merchant logos are longer horizontally than they are vertically, so they were getting squished w/o `object-fit: cover`. We could also use contain or the default, but cover seemed the best to me (although there are some risks of the restaurant name being cut off in cover):

Cover
Screen Shot 2020-10-05 at 4 58 43 PM

Contain
Screen Shot 2020-10-05 at 5 56 17 PM

Default
Screen Shot 2020-10-05 at 5 32 09 PM

justintmckibben:
I think it should always use SCL's logo: https://www.figma.com/file/I0XPeX6re8t14J4nZOvVNp/Send-Chinatown-Love?node-id=4095%3A0 and never use the hero image except for regular gift cards on the second page

justintmckibben (approved):

chrisng93:
@justintmckibben got it, i changed the logo to be the SCL logo at all times:

Screen Shot 2020-10-05 at 6 14 18 PM
Screen Shot 2020-10-05 at 6 14 29 PM
Screen Shot 2020-10-05 at 6 14 39 PM

src/pages/VoucherRedemption/StoreBanner.tsx
chrisng93: never mind probably don't need to get any input since we're using scl logo across the board here

justintmckibben:
Is it possible to add more padding-left to the `?` in both redemption flows to match the mocks? https://www.figma.com/file/I0XPeX6re8t14J4nZOvVNp/Send-Chinatown-Love?node-id=834%3A5

chrisng93:
@justintmckibben couldn't find web mocks, but gave them a little more left margin

Screen Shot 2020-10-06 at 9 35 24 PM
Screen Shot 2020-10-06 at 9 35 03 PM

src/pages/VoucherRedemption/StoreBanner.tsx
jacoblb64: The SCL logo is just the fallback right? Cover looks like the right option here, I think your instinct is right! The only time I imagine that could be weird is if we have the logo as a photo of the owners--but I do think we've generally tried to avoid that anyway.

PR #342, opened 2020-10-07
UI for fiscal sponsor

danthaman44:
Screen Shot 2020-10-06 at 9 08 03 PM
Screen Shot 2020-10-06 at 9 31 09 PM


jacoblb64 (approved):
This looks great to me! How does it look on different widths? Also, nice use of passing a variable to the translation--I didn't know you could do that!

PR #347, opened 2020-10-19
Merchant dashboard v2 base functionality

chrisng93:
Merchant dashboard v2 base functionality. This should almost bring us on par with what we have today. There are still a bunch of todos here though, mainly around the refresh/print buttons, edit functionality in the table, the custom search bar, and the hide GAM checkbox.

Ticket https://trello.com/c/sqaKfnSV/605-merchant-dashboard-20

Full screen
Screen Shot 2020-10-19 at 6 54 46 PM

Smaller screen
Screen Shot 2020-10-19 at 7 08 10 PM

Mobile (table is kind of bad but not sure if there's anything we can really do there other than reduce # of cols)
Screen Shot 2020-10-19 at 7 08 37 PM


src/components/App/App.tsx
```diff
@@ -118,9 +121,12 @@ const App = () => {
           >
             <PassportRedemption screen={ScreenName.Claim} />
           </Route>
-          <Route path="/:seller_id/dashboard/:secret_id">
+          <Route exact path="/:seller_id/dashboard/:secret_id">
             <MerchantVoucherDashboard />
           </Route>
+          <Route exact path="/:seller_id/dashboard/:secret_id/v2">
```
chrisng93, line 127: could do a url param here as well but the new url was easier

jacoblb64: This is fine for now

jacoblb64 (commented):
Looks good to me so far! I don't think we usually follow a pattern of checking in stuff like this before it's ready--not that there's anything wrong with it! We can discuss in our meeting this evening.

jacoblb64 (approved):

chrisng93:
@jacoblb64 I actually updated the PR to include more functionality so that it's at least presentable

jacoblb64 (approved):
Looks good! Much cleaner and way better organized 👍

PR #349, opened 2020-10-19
Removing SCL Food Crawl from the navbar

jacoblb64:
- Food crawl's over, we can remove the link from the top navbar
- Already removed from the weebly site


wilsonj806 (approved):
Looks good to me!

PR #356, opened 2020-10-21
Wilson/prettier precommit hook

wilsonj806:
Add a pre-commit hook to run Prettier with so that it just runs only on whatever files are staged.

It runs via the [Lint Staged](https://www.npmjs.com/package/lint-staged) and [Husky](https://typicode.github.io/husky/#/) package along with some extra config in `package.json`. It runs Prettier using the config at the root of the repo, so there shouldn't be weird collisions when someone runs `yarn format`


justintmckibben (approved):
wow thanks for doing this!

jacoblb64 (approved):
So happy about this :) Thanks @wilsonj806!

wilsonj806:
@jacoblb64 @justintmckibben ye no prob!

PR #358, opened 2020-10-22
Edit capabilities for gift card amount

chrisng93:
Edit capabilities for gift card amount. "Date last used" is implicitly updated to today's date whenever the gift card amount is updated.

Screen Shot 2020-10-22 at 2 11 02 PM

Screen Shot 2020-10-22 at 2 10 26 PM

Depends on https://github.com/sendchinatownlove/ruby/pull/264

### DIFF
--- a/src/utilities/api/types.ts
+++ b/src/utilities/api/types.ts
@@ -30,8 +30,10 @@ export type GiftCardDetails = {
   email: string;
   expiration: string | null;
   gift_card_id: string;
-  // The latest gift card amount's created_at date. This denotes the last
-  // time that the gift card was updated.
+  /**
+   * The latest gift card amount's created_at date. This denotes the last
+   * time that the gift card was updated.
+   */
   last_updated: string;
   latest_value: number;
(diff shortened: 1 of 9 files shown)


src/utilities/api/types.ts
```diff
@@ -29,6 +29,10 @@ export type GiftCardDetails = {
   created_at: string;
   email: string;
   expiration: string | null;
+  gift_card_id: string;
+  // The latest gift card amount's created_at date. This denotes the last
```
jacoblb64, line 33: nit: you can make this a tsdoc comment with /** .. */, so it'll appear for this variable on hover everywhere :)

jacoblb64 (approved):
Looks good man! We synced offline about this simpler editing UI that implicitly sets the redemption date to the current day. I think this makes a lot of sense now.

src/utilities/api/types.ts
chrisng93: oh nice. done

PR #359, opened 2020-10-23
Updated GAM second step copy

danthaman44:
Screen Shot 2020-10-22 at 11 27 27 PM
Screen Shot 2020-10-22 at 11 27 35 PM

### DIFF
--- a/src/locales/cn/translation.json
+++ b/src/locales/cn/translation.json
@@ -120,7 +120,7 @@
       "step1": "第一步：",
       "step1Text": "捐款给爱心餐活动",
       "step2": "第二步：",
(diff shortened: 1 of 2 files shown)


nanxiy:
can we leave off the "." at the end so it's consistent with the other steps :) 

does this also update on the GAM page?

danthaman44:
> can we leave off the "." at the end so it's consistent with the other steps :)
> 
> does this also update on the GAM page?

Updated to remove the period. Yes it updates on the GAM page too, updated the screenshots :)

justintmckibben (approved):

PR #360, opened 2020-10-24
Created basic page structure for Light Up Chinatown

danthaman44:
Screen Shot 2020-10-23 at 11 50 13 PM


src/components/App/App.tsx
```diff
@@ -143,6 +147,9 @@ const App = () => {
           <Route path="/gift-a-meal-home">
             {returnComponent(Page.GiftAMeal)}
           </Route>
+          <Route path="/light-up-chinatown">
```
justintmckibben, line 150: I wonder if we want this to be dynamic, determined by an endpoint that lists all of the "projects"

src/components/LightUpChinatown/LightUpChinatownPage.tsx
```diff
@@ -0,0 +1,98 @@
...
+  font-style: normal;
+  font-weight: bold;
+  font-size: 32px;
+  line-height: 44px;
+  letter-spacing: 0.02em;
+  color: #FFFFFF;
+  position: absolute;
+  top: 35%;
+  left: 50%;
+  transform: translate(-50%, -50%);
```
justintmckibben, line 66: Why not just use flexbox center to center this?

src/components/LightUpChinatown/LightUpChinatownPage.tsx
```diff
@@ -0,0 +1,98 @@
...
+  z-index: 2;
+`;
+
+const HeaderSubtext = styled.div`
+  font-family: Open Sans;
+  font-style: normal;
+  font-weight: bold;
+  font-size: 24px;
+  line-height: 33px;
+  letter-spacing: -0.01em;
```
justintmckibben, line 76: What's the negative for?

src/components/LightUpChinatown/LightUpChinatownPage.tsx
```diff
@@ -0,0 +1,98 @@
...
+  color: #FFFFFF;
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  z-index: 2;
```
justintmckibben, line 82: What's the z-index for?

src/components/LightUpChinatown/LightUpChinatownPage.tsx
```diff
@@ -0,0 +1,98 @@
...
+  font-family: Open Sans;
+  font-style: normal;
+  font-weight: bold;
+  font-size: 24px;
+  line-height: 33px;
+  letter-spacing: -0.01em;
+  color: #FFFFFF;
+  position: absolute;
+  top: 50%;
+  left: 50%;
```
justintmckibben, line 80: same thing about centering. I'd rather use flexbox or text-align: center if that works

src/components/LightUpChinatown/LightUpChinatownPage.tsx
```diff
@@ -0,0 +1,98 @@
...
+`;
+
+const HeaderText = styled.div`
+  font-family: Open Sans;
+  font-style: normal;
+  font-weight: bold;
+  font-size: 32px;
+  line-height: 44px;
+  letter-spacing: 0.02em;
+  color: #FFFFFF;
```
justintmckibben, line 62: do we not have white defined as a constant anywhere?

justintmckibben (approved):

src/components/App/App.tsx
jacoblb64: That's a good idea justin! Also happy to revisit this work when we get to our next "project" if it means speeding this one up

src/components/LightUpChinatown/LightUpChinatownPage.tsx
jacoblb64: check index.scss, but also isn't this just `white`? Also I totally get the copy/paste from figma ahaha

jacoblb64 (approved):
Looks good so far to me!

Two nonblocking larger points that I think @justintmckibben spoke to as well:
- making it extensible as we build out more projects (i.e. pictures, text, etc from backend)--but that can, and probably should come later
- fixing up the CSS from the figma (not urgent)

src/components/App/App.tsx
danthaman44: I'm totally support making it dynamic once we have a clear design for future projects, and the backend model for "projects" :)

src/components/LightUpChinatown/LightUpChinatownPage.tsx
danthaman44: CSS was taken from the Figma hehe

src/components/LightUpChinatown/LightUpChinatownPage.tsx
danthaman44: To overlay the text on top of the hero image!

src/components/LightUpChinatown/LightUpChinatownPage.tsx
danthaman44: Correct this is just white! I usually just copy and paste the CSS from the Figma

src/components/LightUpChinatown/LightUpChinatownPage.tsx
danthaman44: Flexbox doesn't work super well in this case because the text is overlayed on top of the image, as opposed to being a child :(

PR #362, opened 2020-10-25
Wilson/modal dynamic campaign text

wilsonj806:
Update GAM modal to render the campaign distributor off of API data rather than hard coded data. It fetches the distributor for the campaign that expires the soonest(didn't use the distributor that was included with the `getSeller()` API utility function since it didn't include distributor URL)

### DIFF
--- a/src/components/ModalBuyMealNew/ModalBuyMealPool.tsx
+++ b/src/components/ModalBuyMealNew/ModalBuyMealPool.tsx
@@ -50,8 +50,7 @@ export const Modal = (props: Props) => {
       type: ModalPaymentConstants.SET_AMOUNT,
       payload: String(totalMealPrice),
     });
-    // eslint-disable-next-line
-  }, []);
+  }, [dispatch, totalMealPrice]);
 
   return (
     <form data-testid="ModalBuyMeal">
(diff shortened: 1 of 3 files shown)


src/components/ModalBuyMealNew/ModalBuyMealPool.tsx
```diff
@@ -45,8 +48,34 @@ export const Modal = (props: Props) => {
...
+        <a
+          href={campaignDistributor.website_url}
+          target="_blank"
+          rel="noopener noreferrer"
+          className={styles.link}
+        >
+          {' '}
+          {campaignDistributor.name}
+        </a>
+        { ' '}
```
jacoblb64, line 76: nit: extra space here?

src/components/ModalBuyMealNew/ModalBuyMealPool.tsx
```diff
@@ -59,35 +88,7 @@ export const Modal = (props: Props) => {
...
-        <a
-          href="https://www.cpc-nyc.org/"
-          target="_blank"
-          rel="noopener noreferrer"
-          className={styles.link}
-        >
-          {' '}
-          the Chinese-American Planning Council{' '}
-        </a>
+        { Distributor }
```
jacoblb64, line 91: I think if you want, you can change `Distributor` to a lambda that returns that body, and add it as a component here like `<Distributor></Distributor>` -- up to you

jacoblb64 (approved):
Looks good! Added some non-blocking comments. I don't think we have any plans at the moment to run multiple campaigns for a merchant at any one time, but I think when we do we should update this to show all the distributors for active campaigns instead of just the latest.

wilsonj806:
Gotcha, that shouldn't be too bad. Should probably also check to make sure the campaign is active before listing it's distributor also.

PR #363, opened 2020-10-25
Bruce/updating tests

bvillaroman:
updating all the tests to fit out better use of react hooks
- changes to how we initiate our global react storage for certain components like payment modal to provide better unit testing/better react-hook practices
- changes to some of the i18n to use interpolation
- addition of jest mocks in their own folder instead of within the test files
- better handling of alternative layout designs on seller pages
- more unit tests within seller page


bvillaroman:
when we run: yarn test,  we run all the tests at first. 
a shot of all the tests being run
a shot of the owner panel tests showing the default body values according to the layout design (responsiveness)

PR #366, opened 2020-10-26
Merchant dashboard v2 - save success/error banner states

chrisng93:
Merchant dashboard v2 - save success/error banner states

Success:
Screen Shot 2020-10-25 at 5 27 08 PM

Error (generic):
image (1)

Error (validation):
image


src/pages/MerchantVoucherDashboardV2/VoucherTable.tsx
```diff
@@ -37,21 +43,32 @@ const VoucherTable = ({
...
-      // TODO: Show error banner on error.
+      try {
+        const latestValueCents = parseFloat(latestValue) * 100;
+        await updateVoucher(giftCardId, latestValueCents);
+        fetchData();
+        setEditingRowGiftCardId(null);
+        setErrorType(null);
+        setShowSuccessBanner(true);
+      } catch (error) {
+        if (error.message.indexOf('422') === -1) {
```
jacoblb64, line 64: you can use [`.includes()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) now, no need to use `.indexOf()` anymore. But also, would it be better to check the status code itself instead of the message?

src/pages/MerchantVoucherDashboardV2/Banners.tsx
```diff
@@ -0,0 +1,53 @@
...
+};
+
+type ErrorTypeKeys = keyof typeof ERROR_TYPE;
+export type ErrorTypeValues = typeof ERROR_TYPE[ErrorTypeKeys];
+
+const ERROR_COPY_ENGLISH = {
+  GENERIC:
+    'There was an error saving your changes. Please refresh your page and try again.',
+  // Right now, the only validation error we have on the update voucher
+  // endpoint is if the amount is higher than the current voucher value.
```
jacoblb64, line 20: if you want, you can make these tsdoc comments with `/** ... */` on either/or the `VALIDATION` property and/or the parent object itself.

jacoblb64 (approved):
This looks really good man! Very clear and nice and easy to read. How do they look on a smaller screen (i.e. phone in landscape mode)? Left some small non-blocking changes.

src/pages/MerchantVoucherDashboardV2/VoucherTable.tsx
chrisng93: just was worried a bit about browser compatibility for includes (not sure what browsers our merchants usually use). i can change it if we're not worried

when we throw the error from axios, we don't get a clean http error. it just gets captured somewhere in the message

src/pages/MerchantVoucherDashboardV2/Banners.tsx
chrisng93: oh right, updated this code

chrisng93:
@jacoblb64 here's the error banner on a mobile screen:
image

src/pages/MerchantVoucherDashboardV2/VoucherTable.tsx
jacoblb64: Gotchya on the axios stuff. And no need to worry about browser compatibility--the react builder takes care of that, we specify the browserlist in our package.json

chrisng93: oh sweet, ok i changed it to includes then. much cleaner

PR #374, opened 2020-10-28
Switch merchant dashboard over

chrisng93:
Switch merchant dashboard over, delete old code. The diffs just look weird because I changed the file names. There's no actual code changes


jacoblb64 (approved):
Looks good to me!

PR #378, opened 2020-10-31
centered merchant and highlight containers on ipad

billy-yuan:
Quick CSS fix to center the highlight and merchant containers on iPad (see below; tested with Chrome's developer tools).

I didn't change the sizes of any of the containers so the left and right margins are not the same size, but it should be centered at least. Happy to talk to design about this if its a priority.

**Before** image

**After** 
image


justintmckibben (approved):

jacoblb64 (approved):
Nice and easy!

PR #380, opened 2020-11-02
Extended width of donation bar for mobile to align with above text

billy-yuan:
Extended donation bar so that it lines up with the text above on mobile

**Before**
image

**After**
image


justintmckibben (approved):

justintmckibben:
Thanks for finding these!

PR #382, opened 2020-11-09
Add transaction fee

Kenworthe:
- adds transaction fee to ModalAmount for both fixed & custom amounts.
- fee calculator is built from FeeParams from backend.
- for now, only fetches Square fee, but can accommodate multiple fees if the interaction is refactored from `getFee('square')` to `getFees()`.
  - fee calculation is probably best done on backend though, even if it means calculating fee onChange for custom amount.

To activate the square fee, you'll need to set this on the db: 

image


src/utilities/hooks/ModalPaymentContext/types.ts
```diff
@@ -3,16 +3,20 @@
...
  *
  */
 
-import { BrowsePageSeller } from '../../api/types';
+import { BrowsePageSeller, FeeParams } from '../../api/types';
 
 export type ModalPaymentState = {
-  amount: string;
+  amount: string; // @TODO: refactor amount to be type number
+  feesAmount?: number;
```
justintmckibben, line 18: How come we need this separate amount and can't just sum fees?

justintmckibben (approved):

src/utilities/hooks/ModalPaymentContext/types.ts
Kenworthe: Ah, yeah, I thought about doing that, but decided to keep them separate as an easier way to create a line item for each during the CardDetailsModal step. (One line item for donation amount, second line item for fees).

Another potential way we can approach this is to maybe keep line items (instead of amount/feeAmount) at the context-level (acting like a "cart" as the user advances through the modals) and then just straight up submit those line items to Square on the final step. But I didn't explore that route too much, tbh!

wilsonj806 (approved):
I'm not super familiar with payment flow, but the additions seem sane to me!

src/utilities/hooks/ModalPaymentContext/types.ts
justintmckibben: Yeah it would be nice to have a class of sorts that takes line items and has a calculated attributes of subtotal, fee total, etc.

src/components/ModalPayment/ModalAmount/ModalAmount.tsx
```diff
@@ -75,18 +100,18 @@ export const Modal = (props: Props) => {
         <SelectAmtContainer>
           {buttonAmounts.map((amount) => (
             <button
-              key={amount.text}
+              key={`$${amount}`}
               type="button"
               className={
-                selected === amount.text
+                selectedAmount === `$${amount}`
```
billy-yuan, line 106: Are `amount` and `selectedAmount` are both numbers (but string types)? If so i think `selectedAmount === amount` should be fine otherwise I think the buttons won't change color when clicked on. Can double check though

Kenworthe: @billy-yuan yeah good callout; I changed it to compare the amount value now instead of the 'pretty string'. (I mean they're still strings though, but yeah.)

src/utilities/hooks/ModalPaymentContext/types.ts
Kenworthe: Yeah, agreed. And line items should probably be kept higher up, at like the modal payment context-level. Would be good to do in a refactor imo.

PR #383, opened 2020-11-10
Fix infinite network requests bug in passport printouts

wilsonj806:
Low priority but noticed it when Jacob sent a link over.


justintmckibben (approved):
wow what an easy fix lol

PR #396, opened 2020-11-15
Sending Appropriate info to /charges for LUC

jacoblb64:
- Not including a seller ID
- Including the appropriate Project ID
- Including the lucData as metadata (when appropriate)

I know some of this is a little hacky / luc-specific in common code, but I think we have a lot of tech debt in general we'll have to face before our next 'project' :)

Tested with my square dev account -- everything looks good and seems to be working.


aliceluo2048 (approved):

danthaman44 (approved):
LGTM!

PR #398, opened 2020-11-15
Add null safety check to Projects fetch in LUC progress bar

danthaman44:
The LightUpPage won't load on staging/local if there is no backend project. This adds a safety check to make sure it loads


jacoblb64 (approved):
Looks good to me! I added the project to prod so this should be good there too!

justintmckibben (approved):

PR #406, opened 2020-11-16
LUC in nav bar

aliceluo2048:
-  add LUC next to Donate
- remove FOAF from here since they’re no longer donating to us
lucnavagain
image

### DIFF
--- a/src/components/Navbar/NavBar.tsx
+++ b/src/components/Navbar/NavBar.tsx
@@ -47,11 +47,6 @@ const NavBar = (props: Props) => {
 
   const dropdownOptions = [
     { url: '/merchants', translation: 'donate', external: false },
-    {
-      url: '/light-up-chinatown',
-      translation: 'light-up-chinatown',
-      external: false,
-    },
     { url: '/gift-a-meal-home', translation: 'gift-a-meal', external: false },
     {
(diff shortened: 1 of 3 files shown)


src/locales/cn/translation.json
```diff
@@ -16,9 +16,9 @@
       "merchants": "WAYS TO DONATE",
       "waystodonate": {
         "donate": "DONATE",
+        "light-up-chinatown": "LIGHT UP CHINATOWN",
```
justintmckibben, line 19: oops but it looks like it's supposed to be a main navigation: https://www.sendchinatownlove.com/

justintmckibben (approved):
The removal of FOAF looks good!

src/components/Navbar/NavBar.tsx
```diff
@@ -47,17 +47,17 @@ const NavBar = (props: Props) => {
 
   const dropdownOptions = [
     { url: '/merchants', translation: 'donate', external: false },
+    {
+      url: '/light-up-chinatown',
+      translation: 'light-up-chinatown',
+      external: false,
+    },
```
justintmckibben, line 54: I don't think this is supposed to be a part of the dropdown, but is supposed to always be showing to the right of Ways to Donate

image

PR #410, opened 2020-11-17
Fix LUC progress bar

justintmckibben:
The server returns the value in cents
and we assuming it was in dollars


wilsonj806 (approved):
I was just looking at this hah

PR #412, opened 2020-11-17
Remove donation/voucher from min/max error

justintmckibben:


aliceluo2048 (approved):

aliceluo2048:
Are we going to do this for min amount too?

danthaman44 (approved):

justintmckibben:
> Are we going to do this for min amount too?

good catch — amended

PR #413, opened 2020-11-17
Donation progress bar fix

aliceluo2048:
Eg. if percentage raised is 50%
Mobile:
image
Desktop:
image

### DIFF
--- a/src/components/LightUpChinatown/DonationProgressBar.tsx
+++ b/src/components/LightUpChinatown/DonationProgressBar.tsx
@@ -16,8 +16,10 @@ const DonationProgressBar = (props: Props) => {
   return (
     <BarContainer>
       <ProgressBarHeader>
-        {t('lightUpChinatown.progressBarHeader')}{' '}
-        <strong>${centsToDollars(props.raised)} </strong>
+        <strong>
+          {t('lightUpChinatown.progressBarHeader')} $
+          {centsToDollars(props.raised)}
+        </strong>
       </ProgressBarHeader>
       <ProgressBar>
         <Filler style={{ width: `${percentage}%` }} />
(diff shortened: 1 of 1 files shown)


src/components/LightUpChinatown/DonationProgressBar.tsx
```diff
@@ -61,44 +63,39 @@ const Filler = styled.div`
...
+`;
+
+const GoalImage2 = styled(GoalImage)`
+  right: 0;
 `;
 
 const GoalContainer = styled.div`
   width: 100%;
-  postion: relative;
+  position: relative;
```
justintmckibben, line 87: nice

src/components/LightUpChinatown/DonationProgressBar.tsx
```diff
@@ -61,44 +63,39 @@ const Filler = styled.div`
...
   white-space: nowrap;
-  @media (max-width: 920px) {
-    font-size: 8px;
-  }
 `;
 
 const Goal1 = styled.div`
   position: absolute;
-  left: 48%;
+  left: calc(50% - 50px);
```
justintmckibben, line 95: What are these offsets for?

src/components/LightUpChinatown/DonationProgressBar.tsx
```diff
@@ -61,44 +63,39 @@ const Filler = styled.div`
...
   border-radius: 12px;
 `;
-const GoalImage2 = styled(GoalImage1)`
-  left: 98%;
-  @media (max-width: 920px) {
-    left: 92%;
-  }
+
+const GoalImage1 = styled(GoalImage)`
+  left: calc(50% - 25px);
```
justintmckibben, line 78: What's this offset represent?

justintmckibben (approved):
LGTM!

src/components/LightUpChinatown/DonationProgressBar.tsx
aliceluo2048: The offset is so the text is centered around the circle image, as opposed to it starting from the left edge

src/components/LightUpChinatown/DonationProgressBar.tsx
aliceluo2048: It doesn't line up with the width of the % yellow progress bar because of the 25px width of the circle image

justintmckibben: If this were done in scss, I'd say to abstract this into a variable called `$progress-bar-circle-width = 25px` and use it in both places

src/components/LightUpChinatown/DonationProgressBar.tsx
justintmckibben: same here

PR #414, opened 2020-11-17
Mega GAM partially done, code cleanup

chrisng93:
- Correctly sort/order active and inactive GAM campaigns
- Add Mega GAM list item on GAM page
- Type a lot of things
- Little bit of code cleanup

Note that there's a ton of todos still. Will do in followups

Large screen:
Screen Shot 2020-11-16 at 11 09 17 PM

Small screen:
Screen Shot 2020-11-16 at 11 12 02 PM

Ref https://trello.com/c/tZJj4Wkb/725-megagam-active-mega-gam-campaign-list-item

PR #422, opened 2020-11-19
Donation status section for Mega GAM campaigns

chrisng93:
Donation status section for Mega GAM campaigns

Depends on https://github.com/sendchinatownlove/sendchinatownlove.github.io/pull/414

Screen Shot 2020-11-18 at 10 19 15 PM


aliceluo2048:
Could we not have used the existing CampaignProgressBar and change the styling requirements?

chrisng93:
@aliceluo2048 i looked into that but there are enough subtle differences in the design that sharing the code would've made for some confusing code (bar size, font sizes, bar styling, text to the right of % raised, header)

image

image

justintmckibben:
@nanxiy is it not possible to make both bars the same? Either updated to the new look or reverted to the old look? Otherwise we have to maintain two separate components that do the same thing

nanxiy:
> @nanxiy is it not possible to make both bars the same? Either updated to the new look or reverted to the old look? Otherwise we have to maintain two separate components that do the same thing

so here was my logic, I originally had two types of trackers:

1.  larger trackers that exist at half screen or greater (Merchant directory trackers, Mega GAM tracker) as the round trackers
2. and then smaller trackers sit within cards (merchant individual tracker, GAM individual tracker)

but happy to consolidate if it's the same code across the board

justintmckibben:
> > @nanxiy is it not possible to make both bars the same? Either updated to the new look or reverted to the old look? Otherwise we have to maintain two separate components that do the same thing
> 
> so here was my logic, I originally had two types of trackers:
> 
> 1. larger trackers that exist at half screen or greater (Merchant directory trackers, Mega GAM tracker) as the round trackers
> 2. and then smaller trackers sit within cards (merchant individual tracker, GAM individual tracker)
> 
> but happy to consolidate if it's the same code across the board

It's okay to have two different trackers, but could they be the same style only different sizes?

nanxiy:
> > > @nanxiy is it not possible to make both bars the same? Either updated to the new look or reverted to the old look? Otherwise we have to maintain two separate components that do the same thing
> > 
> > 
> > so here was my logic, I originally had two types of trackers:
> > 
> > 1. larger trackers that exist at half screen or greater (Merchant directory trackers, Mega GAM tracker) as the round trackers
> > 2. and then smaller trackers sit within cards (merchant individual tracker, GAM individual tracker)
> > 
> > but happy to consolidate if it's the same code across the board
> 
> It's okay to have two different trackers, but could they be the same style only different sizes?

yes! I can, but just calling out then I will also be updating the merchant trackers as well

justintmckibben:
> > > > @nanxiy is it not possible to make both bars the same? Either updated to the new look or reverted to the old look? Otherwise we have to maintain two separate components that do the same thing
> > > 
> > > 
> > > so here was my logic, I originally had two types of trackers:
> > > 
> > > 1. larger trackers that exist at half screen or greater (Merchant directory trackers, Mega GAM tracker) as the round trackers
> > > 2. and then smaller trackers sit within cards (merchant individual tracker, GAM individual tracker)
> > > 
> > > but happy to consolidate if it's the same code across the board
> > 
> > 
> > It's okay to have two different trackers, but could they be the same style only different sizes?
> 
> yes! I can, but just calling out then I will also be updating the merchant trackers as well

Oh that would be amazing if merchant trackers also looked the same!

nanxiy:
> > > > > @nanxiy is it not possible to make both bars the same? Either updated to the new look or reverted to the old look? Otherwise we have to maintain two separate components that do the same thing
> > > > 
> > > > 
> > > > so here was my logic, I originally had two types of trackers:
> > > > 
> > > > 1. larger trackers that exist at half screen or greater (Merchant directory trackers, Mega GAM tracker) as the round trackers
> > > > 2. and then smaller trackers sit within cards (merchant individual tracker, GAM individual tracker)
> > > > 
> > > > but happy to consolidate if it's the same code across the board
> > > 
> > > 
> > > It's okay to have two different trackers, but could they be the same style only different sizes?
> > 
> > 
> > yes! I can, but just calling out then I will also be updating the merchant trackers as well
> 
> Oh that would be amazing if merchant trackers also looked the same!

@chrisng93 @aliceluo2048 updated all trackers to be rounded but the secondary trackers in cards are obvi smaller. thanks for being flexible :)

chrisng93:
@nanxiy are we able to unify the copy? i think what's annoying about sharing code for the trackers is that the text is subtly different for each use case (see my post above for examples)

nanxiy:
> @nanxiy are we able to unify the copy? i think what's annoying about sharing code for the trackers is that the text is subtly different for each use case (see my post above for examples)

hmm.. I think it's still makes sense to be different for both because the mega-gam is a dollar target where as the individual gam is a meal target. would it be easier to only do the target rather than "__ out of 100 meals" so it's not an always changing number? 
if not, we can brainstorm how to unify that copy

chrisng93:
yeah i was thinking the same thing when i first created this PR - the 2 donation sections are pretty distinctly different in terms of their use cases. it's hard to share code when there are a bunch of small differences, so i'm inclined to keep the mega gam donation section separate from the other gam sections (and the merchant trackers). i think it makes sense to make the progress bars rounded and keep different copy

billy-yuan (commented):
nice. i like how it's easy to switch between the small and large progress bar with one progress bar component

billy-yuan (approved):

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
```diff
@@ -57,18 +55,17 @@ const CampaignProgressBar = ({
...
-            width: `${progressWidth(numContributions, targetAmount)}%`,
-            backgroundColor: progressBarColor,
-          }}
-        ></CurrentProgressBar>
-      </TargetAmountBar>
+      <ProgressBarContainer>
+        <ProgressBar
+          amount={{ current: numContributions, target: targetAmount }}
+          color={progressBarColor}
+          size="SMALL"
```
justintmckibben, line 66: nit: but we should make this an enum

src/components/MerchantsPage/gam/MegaGamProgressBar.tsx
```diff
@@ -0,0 +1,107 @@
+import moment from 'moment';
+import * as React from 'react';
+import { useTranslation } from 'react-i18next';
+import styled from 'styled-components';
+
+import ProgressBar from './ProgressBar';
+
+const PROGRESS_BAR_COLOR = '#CF6E8A';
+
+const centsToDollars = (cents: number) =>
```
justintmckibben, line 10: nit: we should make a util that does this for us across all of the frontend since this is a common thing

src/components/MerchantsPage/gam/MegaGamProgressBar.tsx
```diff
@@ -0,0 +1,107 @@
...
+const Container = styled.div`
+  flex: 1;
+  font-size: 18px;
+  margin-right: 60px;
+  padding: 16px 0;
+  width: 100%;
+`;
+
+const TotalRaised = styled.div`
+  color: #000000;
```
justintmckibben, line 87: Is this the right black? also, there should be a constant somewhere for this black

justintmckibben (approved):
Thanks for doing this! Just some small non-blocking nits

src/components/MerchantsPage/gam/CampaignProgressBar.tsx
chrisng93: done

src/components/MerchantsPage/gam/MegaGamProgressBar.tsx
chrisng93: i can do this in a followup

src/components/MerchantsPage/gam/MegaGamProgressBar.tsx
chrisng93: it's the same as doing `color: black` - it defaults the hex for that to #000000

PR #424, opened 2020-11-20
Use seller_distributor_pairs for gam 

aliceluo2048:
image
front end is broken and isn't showing the seller x distributor title and the learn more about part

we should probably eventually remove distributor_id in campaigns

### DIFF
--- a/src/components/MerchantsPage/gam/CampaignListItem.tsx
+++ b/src/components/MerchantsPage/gam/CampaignListItem.tsx
@@ -94,13 +94,16 @@ const CampaignListItem = (props: Props) => {
           <Description>
             {campaign.description}{' '}
             {distributor && (
-              <a
-                href={distributor.website_url}
-                target="_blank"
-                rel="noopener noreferrer"
-              >
-                Learn more about {distributor.name}.
-              </a>
+              <span>
+                Learn more about{' '}
+                <a
+                  href={distributor.website_url}
+                  target="_blank"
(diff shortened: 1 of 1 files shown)


src/components/MerchantsPage/gam/CampaignListItem.tsx
```diff
@@ -92,13 +94,16 @@ const CampaignListItem = (props: Props) => {
...
-                href={distributor.website_url}
-                target="_blank"
-                rel="noopener noreferrer"
-              >
-                Learn more about {distributor.name}.
-              </a>
+              <span>
+                Learn more about{' '}
+                <a
+                  href={distributor.website_url}
```
justintmckibben, line 100: This can come from the pair as well, but no need to change it now

justintmckibben (approved):

src/components/MerchantsPage/gam/CampaignListItem.tsx
aliceluo2048: Rather change the distributor stuff all in one go than be inconsistent

chrisng93: do one of you want to take this ticket? it encompasses all this work:
https://trello.com/c/wMetd5qi/708-megagam-migrate-campaign-seller-distributor-fields-to-use-the-campaignssellersdistributors-join-table

PR #425, opened 2020-11-23
Wilson/luc hotfixes2

wilsonj806:
Fixes made:
- Update the main `index.scss` so that it imports Open Sans Bold from Google Fonts

- Add conditional background color for the Goal milestone markers in the LUC donation bar and made it dynamic to match the mockups

- Fix LUC donation bar so it matches the width of the content above(the picture is slightly out of date as it's missing the bolded Goal Text that's in the previous 3 images and the milestone marker doesn't have dynamic positioning)

- Update Payment Confirmation modal so that it uses the LUC hero image for LUC instead of the default
- Replace LUC map image with the .gif version

- Fix Luc cost breakdown image responsiveness


src/components/ModalPayment/ModalConfirmation/ModalConfirmation.tsx
```diff
@@ -19,6 +19,9 @@ export type Props = {
   sellerName: string;
 };
 
+const lucHeroImage =
+  'https://storage.cloud.google.com/sendchinatownlove-assets/public/assets/light-up-chinatown/modal-payment-confirm-hero-LUC.png';
```
aliceluo2048, line 23: Why not have the image in our assets?

wilsonj806: Mostly just trying to figure out whether or not we should bundle images or put it in the Storage Bucket instead. Might be over thinking it tbh

src/components/ModalPayment/ModalConfirmation/ModalConfirmation.tsx
```diff
@@ -55,12 +58,15 @@ const ModalConfirmation = (props: Props) => {
     }
   };
 
+  const confirmHeroImage =
+    props.sellerId === 'light-up-chinatown' ? lucHeroImage : confirmationPic;
```
aliceluo2048, line 62: Is there a 'light-up-chinatown' within a sellerId? LUC is a project id @jacoblb64?

wilsonj806: Not right now, see relevant [Trello Card](https://trello.com/c/ltqMmRzu). Should add a TODO comment tho

jacoblb64: Yeah, we need some bigger refactoring which we haven't done yet

jacoblb64 (approved):
Looks all good to me -- but yeah I think we have some refactoring / figuring out to do re:image storage and other things

PR #427, opened 2020-11-23
Add campaign display name to mega GAM UI

chrisng93:
Add campaign display name to mega GAM UI


chrisng93:
this is the same as the one that i diffed out before, but that PR's merge got weird so i diffed it out again

aliceluo2048 (approved):

PR #429, opened 2020-11-25
Fix LUC Map gif

wilsonj806:


justintmckibben (approved):

justintmckibben:
TY for the quick turnaround!

PR #433, opened 2020-12-02
Add fiscal sponsor to mega gam campaign

chrisng93:
Add fiscal sponsor to mega gam campaign. Created a shared FiscalSponsor component to be used by regular and mega gam campaigns

Desktop:
Screen Shot 2020-12-01 at 5 43 06 PM

Mobile:
Screen Shot 2020-12-01 at 5 42 54 PM

### DIFF
--- a/src/components/MerchantsPage/gam/CampaignListItem.tsx
+++ b/src/components/MerchantsPage/gam/CampaignListItem.tsx
@@ -5,11 +5,7 @@ import { tabletScreens } from '../../../utilities/general/responsive';
 import { useTranslation } from 'react-i18next';
 import campaignDefaultImage from '../images/campaign_default.png';
 import { Campaign } from '../../../utilities/api/types';
-import {
-  getDistributor,
-  getSeller,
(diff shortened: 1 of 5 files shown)


nanxiy:
> @nanxiy i ended up making a couple small changes to the design so that we could share code for regular and mega gam campaigns (and also keep consistency)
> 
> 1. the divider b/w image and description is yellow not red
> 2. in mobile view, i kept the fiscal sponsor img
> 
> happy to change either of these things tho

lemme slack you a transparent version of the logo! 
for 1. if we want to streamlilne the divider color let's use HEX F8BA17 since that terrible highlighter color was only used originally because it was Think!Chinatown's logo color.
2. that's fine, is there space to keep the divider line the same width as on web?

chrisng93:
@nanxiy yeah for sure made both those changes:

Screen Shot 2020-12-01 at 6 38 18 PM
Screen Shot 2020-12-01 at 6 38 31 PM

justintmckibben (approved):

PR #439, opened 2020-12-05
Payment modal quick fix

aliceluo2048:
Added margin to fix:
paymentStart

Reverted to original button styling as the confirm button looks like this on mobile:
paymentConfirm
We can make the modal button width back to 100% when we decide to move the back button to the top of the modal


jacoblb64 (approved):
Nice and easy, thanks for the attention to detail :)

PR #440, opened 2020-12-05
Adding maxLength for LUC fields

jacoblb64:
Better late than never
- Middle initial is now forced to 1 character
- Zip code is 5 characters and red when not numeric (not blocking though)
- Reasonable length caps on everything else

I thought about limiting state to 2 characters -- but without validation I'm worried people will start typing "New York" and it'll truncate to "Ne" and suddenly we think people are in Nebraska. So to avoid that keeping it open to abbreviation or full name. Not ideal but we can clean the data later. We already are going to have to do some cleaning as is.

image


.env.development
```diff
@@ -13,8 +13,8 @@ REACT_APP_GA_TRACKING_ID=''
 REACT_APP_STRIPE_PUBLISHABLE_KEY=pk_test_5AByIibLOhR6WHL3Mwnmel3P00zm0pIDrD
 
 # Square publishable keys
-REACT_APP_SQUARE_APPLICATION_ID=sandbox-sq0idb-aUT15n3qbkvAUY2OXQt_fg
-REACT_APP_SQUARE_LOCATION_ID=E4R1NCMHG7B2Y
+REACT_APP_SQUARE_APPLICATION_ID=sandbox-sq0idb-IDT6EwHSQXv2E91PLPaN5Q
+REACT_APP_SQUARE_LOCATION_ID=XPFZV5MHG689R
```
aliceluo2048, line 17: Are we supposed to be committing this?

aliceluo2048 (approved):

.env.development
jacoblb64: Woops -- left that in from testing, thanks for the catch! Reverted!

PR #441, opened 2020-12-05
Wilson/paginated past campaigns

wilsonj806:
- Split off campaign fetching in GAM page so that it's making two requests, one for active campaigns and one for inactive campaigns.
- Update CampaignList item to use updated campaigns API and so it doesn't have to make 2 extra requests
- Small fix to ModalBuyMealPool styling

[Notion Page for screen images](https://www.notion.so/e0fc1a7243594292b27bea6c3094e27b)

Changes to data fetching are based on changes in this [pull request](https://github.com/sendchinatownlove/ruby/pull/304)


src/locales/en/translation.json
```diff
@@ -206,6 +206,11 @@
     },
     "backButton": "Back to Merchants",
     "giftButton": "Gift Meal",
+    "viewMoreButton": "View More",
```
jacoblb64, line 209: do we need these for `cn` translations too?

wilsonj806: I'm actually not sure and included it to be thorough.

src/components/MerchantsPage/gam/GiftAMealPage.tsx
```diff
@@ -21,41 +21,56 @@ const GiftAMealPage = (props: Props) => {
...
-    // Inactive campaigns should show up newest to oldest.
-    setPastCampaigns(past.reverse());
   };
 
   useEffect(() => {
     fetchData();
     // eslint-disable-next-line
   }, []);
 
+  useEffect(() => {
```
jacoblb64, line 56: What's the reason for adding another `useEffect()` call? Can we combine these?

jacoblb64 (approved):
Looks good to me! Cleaner and less calls which is always good :) Left one comment/question but non-blocking.

src/components/MerchantsPage/gam/GiftAMealPage.tsx
wilsonj806: Don't think so, cuz the way I have the effect for fetching past campaigns set up is that it needs some state in order to synchronize, whereas the first one doesn't really depend on anything to synchronize.

PR #442, opened 2020-12-06
Mega gam checkout

billy-yuan:
**Overview**
- I made a new component called `ModalMegaGAM` that is based off of the regular GAM page. Because of many small differences between the two designs, I made a new component as opposed to adding to the regular GAM page
- Updated `ModalCardDetails` so that the `purchaseTypeMessage` is contained in a new component called `TransactionDetails` (line 320). This is because there is conditional rendering depending on whether there's a matched donation, and putting this in a component makes the code a little more readable.
- Added the matched amount `MatchAmount` in the `reducer` file when the action type is `SET_AMOUNT`

**Console log of `makeSquarePayment` props**
- `item_type` is null
- I used `project_id=1` for testing

**Screenshots (desktop & mobile)**
_With donation matching (desktop)_

_With donation matching (mobile)_

_Edge case: Match amount is capped at Target Amount - Current Amount_

_With no donation matching (desktop)_ 

_With no donation matching (mobile)_

**Need to do**

1. Update formatting of meal box so that it is centered (same issue in regular GAM page https://trello.com/c/WkIJCG4n/39-update-meal-formatting)
2. Add boolean for donation match and match amount in back end
3. Update copies once they're ready (particularly, the confirmation page copy)
4. May have to make the bar length in the fiscal sponsor component dynamic so that if the text needs to wrap, then the bar height also adjusts.


src/components/ModalPayment/ModalCardDetails/ModalCardDetails.tsx
```diff
@@ -290,11 +317,65 @@ const ModalCardDetails = ({
...
+        {' '}
+        {purchaseTypeMessage(purchaseType, amount)} of{' '}
+        <b>
+          ${total()} {numberOfMealsText}
+        </b>{' '}
+        to {sellerName}{' '}
+      </span>
+    );
+  };
+
```
billy-yuan, line 373: Lines 362-372 in my diff are the same as lines 352-360 in prod

src/components/ModalPayment/ModalCardDetails/ModalCardDetails.tsx
```diff
@@ -49,12 +52,17 @@ const ModalCardDetails = ({
   costPerMeal,
   nonProfitLocationId,
   campaignId,
```
aliceluo2048, line 54: Do we need this? Especially when we can just do Campaign.Id and `campaignId` doesn't seem to be used anywhere

src/locales/en/translation.json
```diff
@@ -128,7 +137,9 @@
       },
       "details": {
         "donation": "Donation details",
-        "voucher": "Purchase details"
+        "voucher": "Purchase details",
+        "megagam_details_match": "Gift-a-meal donation payment of <0>{{totalPayment}}</0> for a total contribution of <1>{{totalContribution}}</1> with a 1:1 match",
```
aliceluo2048, line 141: `Gift-a-Meal`

src/locales/en/translation.json
```diff
@@ -108,13 +115,15 @@
...
-        "light_up_chinatown": "By proceeding with your transaction, you understand that you are making a donation to the Light Up Chinatown Project. This Project is directed towards supporting Chinatown as a whole, rather than any individual merchant."
+        "light_up_chinatown": "By proceeding with your transaction, you understand that you are making a donation to the Light Up Chinatown Project. This Project is directed towards supporting Chinatown as a whole, rather than any individual merchant.",
+        "mega_gam": "By proceeding with your donation, you understand that you are purchasing a voucher for a local community member to use at one of our participating merchants. Donations may be tax-deductible. Please see the campaign details for more information."
       },
       "header": {
         "completeYour": "Complete your ",
         "paymentInfo": "Payment Information",
         "checkout": "Checkout details",
-        "light_up_chinatown_donation": "Light Up Chinatown Donation"
+        "light_up_chinatown_donation": "Light Up Chinatown Donation",
```
aliceluo2048, line 125: How come Donation is capitalised for LUC but not for mega gam (donation)

src/locales/en/translation.json
```diff
@@ -3,11 +3,17 @@
     "amount": {
       "header": "100% of your support will go to the merchant toward their operating costs. Your transaction is secure, encrypted and directly goes to the owner for sustaining their business.",
       "light_up_chinatown": "100% of donation goes towards permanent light fixtures throughout Chinatown.",
+      "mega_gam":"MegaGAM version: 100% of your support will go to the merchant.",
```
aliceluo2048, line 6: What's "MegaGAM version:"?

src/components/ModalPayment/ModalMegaGam/ModalMegaGam.tsx
```diff
@@ -0,0 +1,333 @@
...
+  };
+  // TODO (billy-yuan): Confirm whether mega-gam campaigns in the future will have donation matching.
+  // If so, we need to add a flag such as isMatching to determine whether to calculate a matchAmount
+  const matchAmount = props.campaign
+    ? calculateMatchAmount(
+        props.campaign.target_amount / 100,
+        props.campaign.amount_raised / 100,
+        donationAmountDollars
+      )
+    : 0;
```
aliceluo2048, line 59: This seems to be used 3 times within the code - maybe you can abstract this and call the function instead?

src/components/ModalPayment/ModalCardDetails/ModalCardDetails.tsx
billy-yuan: `campaignId` is currently used as a prop in the regular GAM payment modal, so if we remove `campaignId` from the card modal, we'd have to update the regular GAM modal too. I do agree there's a bit of redundancy, so I think we can refactor after

src/locales/en/translation.json
billy-yuan: Need to get the latest copy

src/components/ModalPayment/ModalCardDetails/ModalCardDetails.tsx
billy-yuan: Also, one suggestion @justintmckibben had was to create different payment methods within a payment class depending on whether it's a donation, GAM or project (i.e. LUC) as opposed to having only one payment function. That would help ensure only the necessary props are ever passed

PR #444, opened 2020-12-07
Pluralise text for GAM progress bar

aliceluo2048:
Taking advantage of an existing plural Text function for GAM progress 
pluralText


jacoblb64 (approved):
Looks good! I think i18n [also has some pluralization stuff](https://www.i18next.com/translation-function/plurals) we could eventually leverage too

justintmckibben (approved):

PR #445, opened 2020-12-08
Monetary variable renaming

aliceluo2048:
Renamed monetary variable names to include measurement units to avoid confusion between cents and dollars
Some code clean up


billy-yuan (approved):

billy-yuan:
not sure if we want to change all dollar inputs to cents eventually, but this at least makes it clearer

justintmckibben (approved):
thanks so much for doing this!

justintmckibben:
I do wonder if in the future we should consider converting responses from our API into models that can be passed between functions — like these would be a Currency object with a valueInCents() and valueInDollars() method, with a constructor that is built from cents given from the server. Would be neat because we could build shared converters and utils into the models as well

PR #446, opened 2020-12-08
Add dynamic text to first screen of GAM payment flow 

wilsonj806:
forgot it in the original PR for fullscreen payment flow

## Original: 
image

## Updated:
image


curiousYi:
Nit:

"Please enter how many meals from _restaurant_ to gift" 
reads slightly funny to me. 

How about 
"Please enter how many meals from _restaurant_ you would like to gift"

PR #450, opened 2020-12-11
Change luc donationdetail1 text to 14in lantern

wilsonj806:
Also see [trello card](https://trello.com/c/XXlaDe2f)

image


billy-yuan (approved):

aliceluo2048:
@nanxiy The image with the lantern for the min $45 donation needs to be update since it still says 12 inch. We have the text and the lantern as an entire image itself so it's not like we can only just switch the lantern image from 12 to 14 inch.

nanxiy:
> @nanxiy The image with the lantern for the min $45 donation needs to be updated since it still says 12 inch. We have the text and the lantern as an entire image itself so it's not like we can only just switch the lantern image from 12 to 14 inch.

sorry wasn't aware that we were making this change.
desktop image: https://drive.google.com/file/d/1voQCRE5_biAcKBVpcSTieiCnjyIF8rkT/view?usp=sharing
mobile: https://drive.google.com/file/d/1GnB9rMOVJS4Q3dN1gQDGSXG12YrMBvBK/view?usp=sharing

aliceluo2048:
> > @nanxiy The image with the lantern for the min $45 donation needs to be updated since it still says 12 inch. We have the text and the lantern as an entire image itself so it's not like we can only just switch the lantern image from 12 to 14 inch.
> 
> sorry wasn't aware that we were making this change.
> desktop image: https://drive.google.com/file/d/1voQCRE5_biAcKBVpcSTieiCnjyIF8rkT/view?usp=sharing
> mobile: https://drive.google.com/file/d/1GnB9rMOVJS4Q3dN1gQDGSXG12YrMBvBK/view?usp=sharing

Thanks Nanxi - you're the best ^_^

PR #453, opened 2020-12-14
adding base i18n/translation files for passport

bvillaroman:


justintmckibben (approved):
Oh what we're doing translations on the passport?? :eyes:

bvillaroman:
> Oh what we're doing translations on the passport?? 👀

yeah most of it was hardcoded in the website so i thought i might aswell get it out of the way before we do some heavy lifting on the LNY crawl

https://trello.com/c/XXm8gyel/683-add-localization-to-passport-popover

PR #459, opened 2020-12-21
Wilson/fix printed merchant dash

wilsonj806:
Current merchant dashboard uses JavaScript to toggle out elements when printed, but that doesn't work. Updated to use `@media print` to handle hiding stuff and modifying styles to fit.

- [Current Merchant Dash Printout](https://github.com/sendchinatownlove/sendchinatownlove.github.io/files/5725659/Merchant.Dash.12-21-2020.pdf)
- [Updated Merchant Dash Printout](https://github.com/sendchinatownlove/sendchinatownlove.github.io/files/5725711/Updated.Merchant.Dash.pdf)

### DIFF
--- a/src/pages/MerchantVoucherDashboard/VoucherDashboard.tsx
+++ b/src/pages/MerchantVoucherDashboard/VoucherDashboard.tsx
@@ -157,20 +157,18 @@ const VoucherDashboard = ({
           <div className={styles.headerTitle}>Voucher Tracker 礼品券记录</div>
           <div className={styles.headerSubtitle}>{organizationName}</div>
         </div>
-        {!showPrintView && (
-          <div className={styles.actionButtons}>
-            <LoadingButton
-              icon={RefreshIcon}
-              onClick={fetchData}
-              text="Refresh 刷新"
-            />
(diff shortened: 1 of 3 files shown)


justintmckibben (approved):

jacoblb64 (approved):
Ahh this looks like the proper way to do this 😂. Good to know about the `@media` tag. Does the changing the number per page still work?

Also @chrisng93 should definitely take a look

wilsonj806:
> Ahh this looks like the proper way to do this 😂. Good to know about the `@media` tag. Does the changing the number per page still work?
> 
> Also @chrisng93 should definitely take a look

@jacoblb64 Yeah I only found it while randomly looking up media queries a while ago lol. And changing the number per page still works if I manually change the number in the component, but the way it is now, there doesn't seem to be a pretty way to print every gift card at once(the current stuff just rerenders the table but it doesn't affect the printout).

chrisng93:
@wilsonj806 ah thanks for cleaning up this code. this is definitely a clearer way to do this. `@media print` didn't know about this before!

PR #460, opened 2020-12-22
Remove LUC campaign deadline text

wilsonj806:


jacoblb64 (approved):
Are we planning on bringing this back at some point closer to the new ceremony? If so maybe it would be better to hide the elements than delete the code?

Either way not a huge deal and approving.

wilsonj806:
> Are we planning on bringing this back at some point closer to the new ceremony? If so maybe it would be better to hide the elements than delete the code?
> 
> Either way not a huge deal and approving.

Good point, will replace with commented out code and a note instead.

PR #463, opened 2021-01-01
Bruce/cleaning passport rebase

bvillaroman:
better organization for the passport app


src/pages/PassportRedemption/Redemption/RedemptionFooters.tsx
```diff
@@ -45,7 +46,7 @@ export const RedeemRewardsFooter = (props: RedeemRewardsProps) => {
           window.location.href = `/passport/${props.id}/redeem/${props.access_token}/sponsor/${props.selectedSponsor.id}`;
         }}
       >
-        {t('passport.placeholders.redeemNow').toUpperCase()}
+        REDEEM NOW
```
aliceluo2048, line 49: Why did we change to hardcoding? Same with the other text in the file.

jacoblb64 (approved):
Ahh so fresh and clean -- thanks for taking care of this man!

I do also share @aliceluo2048's question about the hardcoded strings instead of using the translation files, but if you've got a good reason, I don't think that's a blocker.

src/pages/PassportRedemption/Redemption/RedemptionFooters.tsx
bvillaroman: ah theres some files that have hardcoded translations because i missed them when i had to do the rebase of dev with the i18n. Will fix then merge

PR #467, opened 2021-01-06
roll back changes since its breaking donate page

billy-yuan:
Reverting changes because the main SCL donate payment page is breaking due to them


billy-yuan:
Added the mega gam logic into a separate if statement so it won't break when the payment type is Donation

justintmckibben (approved):

PR #471, opened 2021-01-10
Update no active GAM campaigns button modal text and set modal footer text

phiyan:
Trello Card: https://trello.com/c/9kWVBoet/780-update-no-active-gam-campaigns-button-modal-text

Screen Shot 2021-01-09 at 9 46 11 PM

I can't figure out why the `<p>` footer text won't take up the entire width in chrome, unfortunately, but I figure it's a minor piece.

Just to clarify the word done:

- Header and body changes to Gift-a-Meal Donation for when there isn't a campaign
- Footer with the disclaimer always shows for all scenarios

It's not great that I have that if statement there, but the work to change the structure, I think, may be a bit more than expected. Do we want to eventually remove business logic from the `Modal` component so the callers just pass in what it wants to render?

### DIFF
--- a/src/components/ModalPayment/ModalAmount/ModalAmount.tsx
+++ b/src/components/ModalPayment/ModalAmount/ModalAmount.tsx
@@ -90,6 +90,9 @@ export const ModalAmount = (props: Props) => {
 
     switch (purchaseType) {
       case ModalPaymentTypes.modalPages.donation:
+        if (sellerName === 'Apex for Youth') {
+          return t('purchase.donationDefault');
+        }
         return t('purchase.donation', { seller: sellerName });
       case ModalPaymentTypes.modalPages.gift_card:
         return t('purchase.voucher', { seller: sellerName });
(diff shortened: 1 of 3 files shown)


src/components/ModalPayment/ModalAmount/ModalAmount.tsx
```diff
@@ -297,3 +304,9 @@ const Header = styled.div`
   font-size: 30px;
   font-weight: 600;
 `;
+
+const Footer = styled.p`
+  position: relative;
+  left: 0;
+  bottom: -100px;
```
justintmckibben, line 311: We try to avoid negative px

justintmckibben (approved):
LGTM! Just a CSS nit

jacoblb64:
@phiyan any updates on this? If you're waiting to fix the negative bottom, we can get this in and follow up on that later!

phiyan:
Sorry, just getting to this now. What's the preferred way to do the layout? The button is using negative bottom positioning, too

jacoblb64:
> Sorry, just getting to this now. What's the preferred way to do the layout? The button is using negative bottom positioning, too

No worries at all -- happy to help. I think generally we try to avoid negative values in our CSS as it's often a code smell for some shortcut being taken or not a good foundation for responsive design. There are obviously exceptions to this, but I think in this case you might be able to achieve the same result with some bottom padding.

There's plenty of heated discussion online, here's just one dev blog I found: https://daverupert.com/2017/03/thoughts-on-negative-margins/

This might actually get into why your `p` isn't spanning the full width either -- I think there's some flex defined in the parent that might be affecting the footer, and may need to place it in a different place in the DOM and/or play with the flex. Happy to work with you on this if you'd like! Feel free to ping me on slack anytime.

phiyan:
Switched to a flex layout for the bottom two pieces, and we can extend it to encapsulate the other pieces too, if wanted. I'm not great at using flex so let me know if I need to change anything!

Screen Shot 2021-01-31 at 10 01 09 PM

src/locales/en/translation.json
```diff
@@ -90,6 +90,8 @@
     "meals": "meals"
   },
   "purchase": {
+    "footer": "Certain Gift-a-Meal campaigns will be conducted in partnership with Apex for Youth, a 501(c)(3) nonprofit organization. If a campaign is denoted as such, contributions will be tax deductible.",
```
jacoblb64, line 93: Apologies for the one-more-thing, but would you mind adding these (in english) to the cn translations file? Otherwise they won't even show up when viewing in chinese :)

jacoblb64 (approved):
Looks great! Thanks for taking the time to refactor and one minor ask!

src/locales/en/translation.json
phiyan: Ah, good catch. My bad.

phiyan:
Before I merge, do I need to know anything or all good?

jacoblb64 (approved):
@phiyan you're good, go ahead and merge!

PR #472, opened 2021-01-12
Swapping the LUC donation component with redirects to other sources

jacoblb64:
Been a while since I've done any front-end stuff, so please feel free to correct me, open to feedback / other ways of doing things. I tried to keep this change as minimal as possible.

Figma: https://www.figma.com/file/I0XPeX6re8t14J4nZOvVNp/Send-Chinatown-Love?node-id=8742%3A182

* creating a new component to swap in place of the DonationSection, which I'm leaving if we need it
* almost a copy of the DonationSection, but with slight new styling and new text
* text in translation files
* removed from the top-level header and into the drop down as requested

Could be improved:
* couldn't quite figure out the right spacing with the centered text for the GoFundMe button and icon
* hover state for the buttons wasn't clear, so tried to follow our trends
* the svg fill color toggling for the hover state

Screen Shot 2021-01-11 at 11 40 21 PM
Screen Shot 2021-01-11 at 11 40 32 PM
Screen Shot 2021-01-12 at 12 13 37 AM


src/components/LightUpChinatown/DonationRedirect.tsx
```diff
@@ -0,0 +1,125 @@
...
+  font-size: 16px;
+  line-height: 22px;
+  letter-spacing: 0.02em;
+  color: #ffffff;
+  @media (${phoneScreens}) {
+    font-size: 13px;
+  }
+`;
+
+const Button = styled.div`
```
wilsonj806, line 107: Probably better to replace the `styled.div` with a `styled.a` and then styling it like a button and feeding it whatever attributes you need.

So rather than handling click through event handler, in application it'd look like:
```js
<Button href='https://www.gofundme.com/f/light-up-chinatown' rel='noopener noreferrer' target='_blank'/>
```

jacoblb64: Good call! I can try that out! I'll see if I can still have the gofundme open in a new tab, but keep our internal one as a redirect.

wilsonj806 (approved):
Looks good!

PR #475, opened 2021-01-25
Bruce/adding track screen updates

bvillaroman:
so these are all based on our old model/static data that i mocked according to jeremys lny models. Will connect it to the backend in a different branch, just wanted to see a static site


src/pages/PassportRedemption/Passport/index.tsx
```diff
@@ -52,9 +45,7 @@ const Passport = (props: Props) => {
   useEffect(() => {
     if (id) {
       getContactInfo(id)
-        .then((contactInfo) => {
-          setShowInstagram(contactInfo.data.instagram);
-          setShowLyftRewardPromo(contactInfo.data.is_eligible_for_lyft_reward);
+        .then((res) => {
```
aliceluo2048, line 48: Are we even using the API result from getContactInfo, if we aren't, why do we have this call?

src/locales/en/translation.json
```diff
@@ -424,19 +447,24 @@
...
       "activeRewardsLink": "View active rewards & giveaways",
       "markUsed": "Mark as used",
       "returnTo": "Return to passport",
       "prizeRedeemed": "PRIZE REDEEMED {{date}}",
-      "readyToRedeem": "READY TO REDEEM",
-      "leftToRedeem": "{{amount}} more stamps until your next reward",
-      "oneLeftToRedeem": "1 more stamp until your next reward"
+      "readyToRedeem": "Thank you for supporting small businesses",
+      "leftToRedeem": "{{amount}} more receipts to your next goal",
+      "oneLeftToRedeem": "1 more receipt to your next goal",
```
aliceluo2048, line 463: i18n has a plural formatter that we can use instead of having to manually switching to the cases ourselves:
https://github.com/sendchinatownlove/sendchinatownlove.github.io/pull/444

bvillaroman: i didnt realize lol ty

src/pages/PassportRedemption/Passport/index.tsx
bvillaroman: for now we dont have any connections to our new backend connected to lny as it is not fully implemented. I have a separate branch that has what the endpoints should look like, just not pushed yet since i dont have the local api setup yet so i ignored any changes to our interactions with the api in this branch

aliceluo2048: Oh I see. Thanks!

PR #484, opened 2021-01-29
Athena/lny

ath-chen:
preliminary pr!!! login + upload screens, hooked up to apis (will be cleaning up tonight + the weekend) 

Screen Shot 2021-01-28 at 10 56 13 PM
Screen Shot 2021-01-28 at 10 56 25 PM
Screen Shot 2021-01-28 at 11 02 19 PM


src/pages/PassportRedemption/SignIn/emailScreen.tsx
```diff
@@ -0,0 +1,126 @@
...
+  };
+
+  const viewUploadScreen = async (email) => {
+    let { data } = await getPassportEmailId(email);
+
+    if (!data) {
+      const { data: newUser } = await createPassportEmailId(email);
+      data = newUser;
+    }
+    push(`/lny-passport/${data.id}/upload`);
```
bvillaroman, line 44: so for me, im not sure why but sometimes the pushing of a new url into history would not always change the screen ,and i realize you would have to use setCurrentScreenView() or however you are changing your screen. I would double check if it both changes the screen and the url just incase/

src/pages/PassportRedemption/SignIn/uploadScreen.tsx
```diff
@@ -0,0 +1,302 @@
...
+                {t('passport.labels.reuploadReceipt')}
+              </ReplaceButton>
+            </>
+          )}
+          <input
+            type="file"
+            id="imageFile"
+            capture="environment"
+            accept="image/*"
+            onChange={(e) => uploadImage(e)}
```
bvillaroman, line 194: you can do onChange={uploadImage} and it would do the samething

src/pages/PassportRedemption/SignIn/uploadScreen.tsx
```diff
@@ -0,0 +1,302 @@
...
+          <InputField
+            name="bill-total"
+            className="indent"
+            type="number"
+            onChange={(e) => {
+              e.preventDefault();
+              setBillTotal(e.target.value);
+            }}
+            value={billTotal}
+            min="10"
```
bvillaroman, line 163: im not sure when you plan to convert the amount to cents, but just making sure you know to use cents instead of dollars

src/pages/PassportRedemption/SignIn/uploadScreen.tsx
```diff
@@ -0,0 +1,302 @@
...
+            accept="image/*"
+            onChange={(e) => uploadImage(e)}
+          />
+        </ButtonTest>
+      </Column>
+
+      <InputContainer className="bottom">
+        <Button
+          primary={true}
+          disabled={!participatingSeller || !receipt || Number(billTotal) < 10}
```
bvillaroman, line 202: same thing here with the dollars/cents conversion

src/pages/PassportRedemption/SignIn/uploadScreen.tsx
ath-chen: good call out, def did not know to make it cents hahahha

src/pages/PassportRedemption/SignIn/uploadScreen.tsx
ath-chen: trueeeee

src/pages/PassportRedemption/SignIn/uploadScreen.tsx
```diff
@@ -0,0 +1,302 @@
...
+
+  const submitReceipt = async () => {
+    if (receipt === null) return false;
+
+    try {
+      const ext = receipt.type.split('/')[1];
+
+      let filename = `${id}-${new Date()
+        .toISOString()
+        .replace(/(:|\.)/g, '-')}.${ext}-${participatingSeller}`;
```
jacoblb64, line 84: could we actually do - id, then participating seller, then date last? might just be easier when we're looking these up later

jacoblb64 (commented):
Looks great to me! I think bruce had some open items so not approving yet

src/pages/PassportRedemption/SignIn/uploadScreen.tsx
ath-chen: im going to put the dollar --> cents conversion when we send it to the b/e. not going to edit it here since this is also tied to the value that is displayed to the user

src/pages/PassportRedemption/SignIn/uploadScreen.tsx
ath-chen: yep! updated

src/pages/PassportRedemption/SignIn/emailScreen.tsx
ath-chen: yep! so this is controlled in the index file (the one i showed earlier) so there's both changing of routes + screen changing

src/pages/PassportRedemption/SignIn/uploadScreen.tsx
```diff
@@ -0,0 +1,302 @@
...
+      const signedUrl = unescape(
+        (await getUploadUrl(filename, receipt.type)).data.url
+      );
+      await sendImage(signedUrl, filename, receipt);
+
+      // upload info + receipt to db
+      let data = await uploadCrawlReceipts(
+        participatingSellerId,
+        id,
+        Number(billTotal) * 100,
```
ath-chen, line 96: @bvillaroman fyi - this is where im doing the dollar --> cents conversion

bvillaroman (approved):
should be good to go (until we break it during our testing party)

ath-chen:
> should be good to go (until we break it during our testing party)

LOL

PR #487, opened 2021-01-31
fixing upload screen bugs for lny passport

ath-chen:
fixing bugs 
- email screen copy 
- disable upload button after initial click 
- disable submission unless it matches an existing user 
- fix regex for file uploads --> only allows alphanumeric 
- clears inputs in case there are any errors

### DIFF
--- a/src/pages/PassportRedemption/SignIn/uploadScreen.tsx
+++ b/src/pages/PassportRedemption/SignIn/uploadScreen.tsx
@@ -32,7 +32,7 @@ const UploadScreen = () => {
 
   // search values + did user select a valid merchant?
   const [sellersSearchBar, setSellersSearchBar] = useState<any>([]);
-  const [selectedSeller, setSelectedSeller] = useState(true);
+  const [isSearchingSellers, setIsSearchingSellers] = useState(true);
 
   // receipt total
   const [billTotal, setBillTotal] = useState('');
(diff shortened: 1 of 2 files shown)


src/pages/PassportRedemption/SignIn/uploadScreen.tsx
```diff
@@ -199,10 +214,19 @@ const UploadScreen = () => {
       <InputContainer className="bottom">
         <Button
           primary={true}
-          disabled={!participatingSeller || !receipt || Number(billTotal) < 10}
+          disabled={
+            !allParticipatingSellers.find(
+              (seller) => seller.name === participatingSeller
+            ) ||
```
ath-chen, line 220: i dont rlly like the way i wrote this check (this is to ensure that the inputted seller is an actual participating seller). but my issue when thinking about this was: initial call for all participating sellers comes back as an array, so i would have to do either a o(n) call and store all sellers in a map initially or i would have to do a o(n) check here. 

i chose to put it here for now since the other would require more refactoring, but idk is there something else i could do? lmk thoughts!!

src/pages/PassportRedemption/SignIn/uploadScreen.tsx
```diff
@@ -118,14 +128,16 @@ const UploadScreen = () => {
             onChange={(e) => {
```
bvillaroman, line 129: i would make this a separate function for readability and also for the reason down below in the next comment

src/pages/PassportRedemption/SignIn/uploadScreen.tsx
bvillaroman: hmmm, i would say make a [disabled, setDisabled] state, and then in a handleChange button ,we could call setDisabled and if the instant that we receive a proper participating seller in the input, set the disabled prop to false. But for now this is good

bvillaroman (approved):

PR #490, opened 2021-01-31
remove onblur

ath-chen:
removed onblur. issue is onblur gets called before onclick, which is why the search bar wasn't populating after selecting a merchant


jacoblb64 (approved):
Thanks for taking care of this so quick!

PR #491, opened 2021-02-02
Add additional delivery icons

wilsonj806:


jacoblb64 (approved):
Looks good! Do we usually follow the same resolution naming conventions?

wilsonj806:
> Looks good! Do we usually follow the same resolution naming conventions?

@jacoblb64 I'm actually not sure, didn't change the names from the original file names that Nanxi attached.

jacoblb64:
> > Looks good! Do we usually follow the same resolution naming conventions?
> 
> @jacoblb64 I'm actually not sure, didn't change the names from the original file names that Nanxi attached.

Ah gotchya, I just took a look myself and we have others with the same convention. As far as I'm aware that's usually a mobile convention for different scaling resolutions, which we don't support 😂. but it's really not a big deal and _definitely_ not a blocker ahaha

PR #492, opened 2021-02-03
Hard-coding LUC amount raised

jacoblb64:
We're seeing a bunch of errors on the back-end as the query for all the amount raised for LUC is huge now! (Kudos to us!). Let's save our backend some stress and hard-code this number as the campaign's closed anyway.

Screenshot of all recent timeout failures on this endpoint:
Screen Shot 2021-02-02 at 8 15 58 PM

(It's also not the most optimal query, but we can fix that separately)


jacoblb64:
This seems to be working now 🤷‍♂️ , and as this page is no longer being visited a lot, I don't think we need this actually

justintmckibben (approved):
lol nice

PR #495, opened 2021-02-15
Wilson/dist tools login view

wilsonj806:
# Overview:
Adds distributor tools login. Route is accessed by visiting: `/distributor/login`

Screens can be found at the attached [Notion page](https://www.notion.so/Dist-Tools-Build-out-a-Login-View-f6b2b3c849de477ea491cc7efa18a592)

Depends on this [Backend PR](https://github.com/sendchinatownlove/ruby/pull/345)

# Tests:
Tests use [axios-mock-adapter](https://www.npmjs.com/package/axios-mock-adapter) to handle mock API calls
- it doesn't handle spying on axios, but it does abstract away API calls and responses

Tests for redirection use a `renderWithRouter` fixture that returns an object with the React Testing Library render function inside.
 
The page should:
- Hit the authentication endpoint and get some login credentials on submission of the email form
- Hit the google auth service if you click on the google sign in button
- Redirect to dashboard if valid token
- Render an error message if the email is invalid


jacoblb64 (approved):
Tests look really cool, looks like a good pattern we can use going forward! Curious to see a screenshot, but looks good to me either way.

wilsonj806:
@jacoblb64 oh my bad, forgot to enable sharing 😅 Screenshots should be visible now

jacoblb64:
> @jacoblb64 oh my bad, forgot to enable sharing 😅 Screenshots should be visible now

@wilsonj806 Thanks for enabling, looks great!

PR #497, opened 2021-02-23
Add arrow buttons for Store Gallery Tab Images Modal

curiousYi:
[Ticket](https://trello.com/c/gdtAR4zu/498-add-arrow-buttons-for-the-gallery-images-so-you-can-click-through-them-in-modal-view)

Description:
Adds bumper buttons for a store's gallery modal so the user 
can scroll through the images without having to exit the Modal.


wilsonj806 (approved):
Looks good!

jacoblb64 (approved):
Looks great to me, thanks for taking care of this! Looks like the arrows in the circles might not be perfectly center, but maybe my 👀  are playing tricks on me. Either way let's get this in and we can tweak later if need be!

PR #499, opened 2021-02-28
[RFC] Live metrics frontend

mllee:
Built out a basic frontend similar to the main metrics component we have on weebly right now. I haven't worked on querying the backend yet -- will likely need some code pointers for that.

I have a few open questions for the frontend, hence the RFC:
1. Do we have some set of standard fonts/colors/sizes we use? I couldn't find any, hence the clowny looking fonts/sizes/colors. Will need to substitute my best guesses for something that matches.
2. How does routing work on this app? I tried my best to mirror some existing routes, but couldn't figure it out. Any pointers? I imagine i'd have this point to sendchinatownlove.com/livemetrics, and then pop that into an iframe for the weebly page.

Original: 
image

Reconstructed:
image


jacoblb64:
Hey @mllee, thanks for taking a first crack at this! Looks good so far and let me try to answer your two questions:

1. We do have a general design guide / palette on Figma [here](https://www.figma.com/file/I0XPeX6re8t14J4nZOvVNp/Send-Chinatown-Love?node-id=484%3A0), I think there's even a (slightly outdated) [design for the landing page where you can grab some colors / styling as well](https://www.figma.com/file/I0XPeX6re8t14J4nZOvVNp/Send-Chinatown-Love?node-id=127%3A1026). Let me know if that helps.
2. The whole react app is hosted on `localhost` well, locally, and on `merchant.sendchinatownlove.com` in production. Routing from there is handled in App.tsx which it looks like you added your path too. I'm not the most familiar with how to host content to be embedded into an iframe, so maybe we can work together on the best way to do that. What else were you curious about?

Feel free to ping me on slack if you'd like to chat there too!

jacoblb64:
Closing this as it's now live! @stanzheng shepherded this through :)

PR #505, opened 2021-03-26
Remove LUC from Dropdown

aliceluo2048:
I miss Eng


jacoblb64 (approved):
Wait we never did this? Thanks for taking care of it!

PR #509, opened 2021-04-08
addressing confirm button not re-enabling on card details edit

kaiserfenix:
### [Description of changes]

- adding some translations that seem to be without parity (i have missed others, i'm sure)
- adding translate-able text to the "Terms and Conditions" piece, in EN + CN
- ??? (WIP)


src/components/ModalPayment/ModalCardDetails/ModalCardDetails.tsx
```diff
@@ -471,13 +483,16 @@ const ModalCardDetails = ({
             <CheckboxContainer>
               <Checkbox
                 value="checkedA"
-                inputProps={{ 'aria-label': 'Checkbox A' }}
+                inputProps={{ 'aria-label': 'Terms and Conditions' }}
```
kaiserfenix, line 486: were these two left as "checkbox a" and "b" on purpose? or just a lack of a better name for them

src/components/ModalPayment/ModalCardDetails/ModalCardDetails.tsx
```diff
@@ -421,6 +426,12 @@ const ModalCardDetails = ({
             cardNonceResponseReceived={cardNonceResponseReceived}
             formId="SPF"
             apiWrapper=""
+            inputEventReceived={() => {
+              console.log('??', checkFormValidity());
```
aliceluo2048, line 430: merp

src/components/ModalPayment/ModalCardDetails/ModalCardDetails.tsx
jacoblb64: This looks like something pretty old that we just never changed. Looks like the last person to touch these was @wilsonj806 and they were made by @ath-chen, either of you have any input? While we're changing the aria-labels, should we change the values to something more meaningful too?

src/locales/cn/translation.json
```diff
@@ -56,7 +56,9 @@
   },
   "purchase": {
     "donation": "捐款 - {{seller}}",
-    "voucher": "购买礼品券 - {{seller}}"
+    "voucher": "购买礼品券 - {{seller}}",
+    "donation_to": "Donation to {{seller}}",
```
jacoblb64, line 60: nit: I think we usually use camelCase for these

src/utilities/testing/render.tsx
```diff
@@ -1,13 +1,13 @@
 jest.mock('../../utilities/api/interactionManager');
 jest.mock('react-router-dom', () => ({
-  ...jest.requireActual('react-router-dom'),
+  ...(jest.requireActual('react-router-dom') as typeof ReactRouterDom),
```
jacoblb64, line 3: I'm cool to check these in as a fix for that error -- but do want to check with @bvillaroman. We can also get these in separately if we're feeling low-confidence about them.

src/components/ModalPayment/ModalCardDetails/ModalCardDetails.tsx
```diff
@@ -421,6 +426,12 @@ const ModalCardDetails = ({
...
+              if (checkFormValidity()) {
+                setCanSubmit(true);
```
jacoblb64, line 432: any reason not to do `setCanSubmit(checkFormValidity())` ?

src/components/ModalPayment/ModalCardDetails/ModalCardDetails.tsx
```diff
@@ -268,13 +269,17 @@ const ModalCardDetails = ({
   };
 
   useEffect(() => {
-    setCanSubmit(
+    setCanSubmit(checkFormValidity);
```
jacoblb64, line 272: does `checkFormValidity` need its `()`? functions as variables always confuse me.

jacoblb64 (approved):
Left some minor comments but all looks great to me -- none are blockers. Thanks for taking care of this so quickly!

On an unrelated note... I wonder if we need somewhere that actually has these "terms and conditions" we reference 🤔

src/components/ModalPayment/ModalCardDetails/ModalCardDetails.tsx
jacoblb64: This might fix your build issue -- not 100% sure

src/components/ModalPayment/ModalCardDetails/ModalCardDetails.tsx
ath-chen: i think these were just never updated haha, yeah the value should be changed as well. thanks for catching and updating!

src/locales/cn/translation.json
kaiserfenix: yeah; i think you're right. but this is already used in existing places.  it would be too out of scope to adjust this. we can (and should) make a ticket to standardize the translations to bring it to parity.

src/components/ModalPayment/ModalCardDetails/ModalCardDetails.tsx
kaiserfenix: i'm not sure what's using this down stream, but i changed the label for now. we can (and should) create a ticket to address the value. i did a search-all for "checkedA" couldn't find any usages. unless you can tell me @ath-chen that it's not "used" anywhere yet. then i can updated in this PR with minimal scope creep

src/components/ModalPayment/ModalCardDetails/ModalCardDetails.tsx
kaiserfenix: no; not particularly. can be condensed. good catch

src/components/ModalPayment/ModalCardDetails/ModalCardDetails.tsx
kaiserfenix: ah; that seems likely. will commit it.

jacoblb64 (approved):
Discussed offline, while this isn't necessarily the ideal solution, this is already a lot better at unblocking folks and we should get this in! Will likely create some follow up item(s) to rework this flow

src/utilities/testing/render.tsx
bvillaroman: ah damn typescript lol. Yeah this looks good for me

kaiserfenix: thanks bruce

PR #510, opened 2021-04-13
#702 fix show queens restaurants under queens

stanzheng:
## [Ticket](insert_trello_card_link_here)

https://trello.com/c/8cuqL3Ps/702-fix-boc-bocs-location-filtering
### [Description of changes]
fix filtering for bocboc

image

### DIFF
--- a/src/components/MerchantsPage/MerchantsPage.tsx
+++ b/src/components/MerchantsPage/MerchantsPage.tsx
@@ -76,13 +76,14 @@ const MerchantsPage = (props: Props) => {
           store!.locations.length > 0 && store!.locations[0].city === type
       );
       if (type === 'Floral Park') {
-        const flushing = sellers.filter(
+        const queens = sellers.filter(
           (store: any) =>
             store!.locations.length > 0 &&
             (store!.locations[0].city === 'Flushing' ||
-              store!.locations[0].city === 'Queens')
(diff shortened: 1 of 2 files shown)


wilsonj806 (approved):
looks good!

src/components/MerchantsPage/MerchantsPage.tsx
```diff
@@ -76,12 +76,14 @@ const MerchantsPage = (props: Props) => {
           store!.locations.length > 0 && store!.locations[0].city === type
       );
       if (type === 'Floral Park') {
-        const flushing = sellers.filter(
+        const queens = sellers.filter(
```
jacoblb64, line 79: This logic confuses me, why is if (floral park) the top level, and does that need to be in the or statement? In any case I trust that this works 😬 thanks for hopping on this!

jacoblb64 (approved):

src/components/MerchantsPage/MerchantsPage.tsx
stanzheng: so it seems like floral park is the default as it was the first so it was always classified as "queens"

jacoblb64: Ahhh makes sense, thanks for the clarification!

PR #514, opened 2021-04-18
Sticky navbar

slants3:
Changing the NavBar to be sticky during mobile viewport sizes.

Two things to note:
- It looks like for whatever reason styled-components isn't updating based on prop changes here - https://github.com/sendchinatownlove/sendchinatownlove.github.io/blob/5217c9e7c7ab5079d21d2656ee5086c0bb277419/src/components/Footer/Footer.tsx#L50 As a result, I opted for a fairly janky solution of changing the `NavLink` margins to padding instead and hiding the footer behind a white background. Not perfect, but I couldn't figure out why the the footer display refused to change to None on props change
- Added a classname specifically for the Logo to target it for CSS purposes

### DIFF
--- a/src/components/Navbar/NavBar.tsx
+++ b/src/components/Navbar/NavBar.tsx
@@ -230,7 +230,7 @@ const HeaderContainer = styled.header`
     position: sticky;
     position: -webkit-sticky;
     top: 0;
-    z-index: 10;
+    z-index: ${theme.maxzIndex};
     background-color: white;
     border-bottom: 1px solid #f2f2f2;
     width: 100%;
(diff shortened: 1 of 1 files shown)


jacoblb64 (approved):
Damn killin it, thanks for taking care of this! Would be curious to see a screenshot with the padding and all, but I'm sure it's fine!

src/components/Navbar/NavBar.tsx
```diff
@@ -220,6 +220,15 @@ const HeaderContainer = styled.header`
     `
     margin-top: 40px;
   `}
+
+  @media (max-width: 600px) {
```
jacoblb64, line 224: I think we have directives for standard widths, e.g. `tablet-width` somewhere if you want to grab that

jacoblb64:
[Make header / burger menu stationary on landing page for mobile](https://trello.com/c/8rSv0hqT/806-make-header-burger-menu-stationary-on-landing-page-for-mobile)

slants3:
Here's it what it looks!

image

There's a bug with a footer that I have to fix - https://github.com/sendchinatownlove/sendchinatownlove.github.io/blob/5217c9e7c7ab5079d21d2656ee5086c0bb277419/src/components/Footer/Footer.tsx#L50

It looks like we expect the menu opening to hide the footer but as far as I can tell wen ever actually hide the footer and it's causing collision with the navbar when you open the hamburger

image

Investigating...

PR #523, opened 2021-05-05
fix: navbar open menu no longer broken for some iPhone users

jacoblb64:
### [Description of changes]
- conditionally applying sticky logic only when the menu is closed

Screen Shot 2021-05-05 at 7 41 42 PM

- tested on an iPhone SE v1 on iOS 14.3

### Other notes
- I think all our conditionality in the inline `styled` CSS is working... we should look into it and fix sometime


jacoblb64:
[Bugfix for sticky navbar on iOS](https://trello.com/c/pt5EGq1I/838-bugfix-for-sticky-navbar-on-ios)

slants3 (approved):
Incredible!

PR #528, opened 2021-05-06
Wilson/split merchant dir req

wilsonj806:
## [Ticket](https://trello.com/c/eu6uDsZQ)

Refactor so rendering of merchant's directory isn't blocked by missing data from the backend. Add loaders while the backend does its thing for more good UX.
https://user-images.githubusercontent.com/12119553/117373159-b2070a80-ae98-11eb-8c98-7078d5da5164.mp4

https://user-images.githubusercontent.com/12119553/117373172-b9c6af00-ae98-11eb-9a0c-13f6ff763aa2.mp4

yerr

### DIFF
--- a/src/components/Loader/LinearLoader.tsx
+++ b/src/components/Loader/LinearLoader.tsx
@@ -13,9 +13,9 @@ const Loader = () => {
 export default Loader;
 
 const LinearLoader = styled(LinearProgress)`
-  height: 20px !important;
+  height: 16px !important;
   width: 100%;
-  border-radius: 10px;
+  border-radius: 8px;
   background-color: #e6c5ca !important;
   & div {
     background-color: #ab192e !important;
--- a/src/components/Loader/LinearLoader.tsx
+++ b/src/components/Loader/LinearLoader.tsx
@@ -1,0 +1,37 @@
+import React from 'react';
+import LinearProgress from '@material-ui/core/LinearProgress';
+import styled from 'styled-components';
+
+const Loader = () => {
+  return (
+    <LoaderContainer>
+      <LinearLoader />
+    </LoaderContainer>
+  );
+};
+
+export default Loader;
+
+const LinearLoader = styled(LinearProgress)`
+  height: 20px !important;
+  width: 100%;
+  border-radius: 10px;
+  background-color: #e6c5ca !important;
+  & div {
+    background-color: #ab192e !important;
+  }
+`;
+
+const LoaderContainer = styled.div`
+  width: 100%;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  height: 100%;
+`;
+
+export const LoaderFillerContainer = styled.div`
+  width: 100%;
+  height: 74vh;
+`;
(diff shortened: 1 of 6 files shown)


wilsonj806:
[Split off merchant page API requests into 2 separate requests](https://trello.com/c/eu6uDsZQ/831-split-off-merchant-page-api-requests-into-2-separate-requests)

src/utilities/api/interactionManager.ts
```diff
@@ -500,3 +501,9 @@ export const getVoucherMetadata = async () =>
 
     .then((res) => res)
     .catch((err) => err);
+
+export const getTotalContributions = async () =>
+  axios
+    .get(totalContributions)
+    .then((res) => res)
+    .catch((err) => err);
```
jacoblb64, line 509: I know we do this a lot, but I'm personally a fan of not catching and doing nothing here, but rather catching in the underlying consuming component and actually doing something about it -- not sure what that would be in this case though

src/components/Loader/LinearLoader.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const Loader = () => {
+  return (
+    <LoaderContainer>
+      <LinearLoader />
+    </LoaderContainer>
+  );
+};
+
+export default Loader;
```
jacoblb64, line 13: nit: exported component objects should match the file name

maybe name the internal `styled(LinearProgress)` something different?

src/components/MerchantsPage/MerchantsGridContainer.tsx
```diff
@@ -0,0 +1,29 @@
...
+import styled from 'styled-components';
+
+const MerchantsGridContainer = ({ loadingSellers, children }) => {
+  return <Container loadingSellers={loadingSellers}>{children}</Container>;
+};
+
+export default MerchantsGridContainer;
+
+const Container = styled.div`
+  display: ${({ loadingSellers }: { loadingSellers: boolean }) =>
```
jacoblb64, line 11: does this inline syntax in the static css work?

src/components/MerchantsPage/MerchantsPage.tsx
```diff
@@ -34,43 +35,52 @@ const INITIAL_CONTRIBUTIONS: ContributionsType = {
...
     );
+    setLoadingSellers(false);
   };
 
   useEffect(() => {
     fetchData(i18n.language);
   }, [i18n.language]);
 
+  useEffect(() => {
+    const asyncFetch = async () => {
```
jacoblb64, line 66: do we need another `useEffect()` hook, or can we just declare this function in the larger component and call it from the same hook? or even just place the call inside `fetchData()`? If it needs to be a separate function, let's name it something accordingly

jacoblb64 (approved):
Looks super fancy man! This is honestly really awesome. Left some nitpicky comments about names and such, but nothing is blocking to me

src/utilities/api/interactionManager.ts
wilsonj806: Yeah agreed, it's janky that we just return an error and nothing else. We'd have to go through all endpoints we call and see what needs to be updated on the UI side to reflect the error if we do so though so idk.

src/components/MerchantsPage/MerchantsPage.tsx
wilsonj806: I prefer having separate effects mostly for clarity, one useEffect does x while the other does y rather than make one useEffect do everything. Although I'll update the function names because that's no bueno.

PR #529, opened 2021-05-07
Wilson/file management smilew

wilsonj806:
## Changes
- Move a bunch of page components from /src/components to /src/pages.
- Small fix to merchant page
- Swap out API endpoint in the development env file to use staging by default
- Add placeholder merchant card image size so the card doesn't need to grow to final size on image load
- Refactor App so that the "/" redirects to the merchant directory instead of just 404

## Merge after Eng Sync Monday!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!


jacoblb64 (commented):
Biiiig refactor, love it! I agree, let's merge after the eng sync so we can alert others -- this could create merge conflicts galore

stanzheng (approved):

PR #530, opened 2021-05-07
Sticky Navbar fix with media query

slants3:
https://trello.com/c/pt5EGq1I/838-bugfix-for-sticky-navbar-on-ios

Implementing previous fix using a conditional to apply `position: sticky` - this time with media query so this only occurs on small screens.

### DIFF
--- a/src/components/Navbar/styles.module.scss
+++ b/src/components/Navbar/styles.module.scss
@@ -1,0 +1,10 @@
+.sticky {
+  @media (max-width: 599px) {
+    top: 0;
+    position: sticky;
+    position: -webkit-sticky;
+    position: -moz-sticky;
+    position: -o-sticky;
+    position: -ms-sticky;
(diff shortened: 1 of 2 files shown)


src/components/Navbar/styles.module.scss
```diff
@@ -0,0 +1,10 @@
+.sticky {
+  @media (max-width: 599px) {
```
jacoblb64, line 2: how does this look on tablet / different screen sizes? Let's test the full gamut this time 😂

jacoblb64 (approved):

src/components/Navbar/styles.module.scss
slants3: This only applies on mobile viewports, anything greater won't apply the `position: sticky`

stanzheng (approved):

PR #522, opened 2021-05-03
Revert "addressing confirm button not re-enabling on card details edit"

stanzheng:
Reverts sendchinatownlove/sendchinatownlove.github.io#509

PR #531, opened 2021-05-07
fixing purchase flow to prevent double clicking

bvillaroman:
adding loading animation for payment button when interacting with square api when authorizing payment

## [Ticket](insert_trello_card_link_here)
https://trello.com/c/PGyHxwIB/837-high-priority-fix-purchase-flow-to-prevent-double-clicking
### [Description of changes]
added loading animation and loading toggling whena nonce is being made and being created

### DIFF
--- a/src/components/ModalPayment/Modal.tsx
+++ b/src/components/ModalPayment/Modal.tsx
@@ -39,10 +39,10 @@ export const Modal = (props: Props) => {
   const dispatch = useModalPaymentDispatch(null);
 
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


bvillaroman:
> Looks good and all makes sense to me!
> 
> I was wondering if any of the stuff we reverted here can be brought back, but we can also address that separately
> #522
> 
> Approved!

Nope, it should have the original changes in tact. All that would happen is show apading animation to when the button is complete. If the post req to create a nonce/auth the card fails then the loading anime stops

PR #532, opened 2021-05-07
make faster by caching results so its not refreshed between merchant …

stanzheng:
…navigation trello #842

### [Description of changes]

make even faster by persisting network results between navigation


wilsonj806 (approved):
Ngl this seems like massive overkill BUT WE HAVE TO GO EVEN FURTHER BEYONDDDDDDDDDDDDDDDD

jacoblb64:
Ahaha caching on both sides, love it! 🚀 🚀 🚀

PR #535, opened 2021-05-10
Remove extra parenthesis in Merchant Page

aliceluo2048:
Before
image
After
after


wilsonj806 (commented):
Oh I actually hotfixed this in this [PR](https://github.com/sendchinatownlove/sendchinatownlove.github.io/pull/529) but we're merging it tonight because there's a lot of other breaking changes in it.

PR #538, opened 2021-05-23
Footer - Update social media icons

carolinexsun:
### [Description of changes]
* Replaced font-awesome icons with image icons
* Updated social media links in consts.ts
* Updated responsive behavior to better match the homepage, but not an exact match
footer-desktop

footer-mobile

Could use some feedback on the following:
* Better way to reference the various new images in consts.ts?
* Unsure if Footer component is actually utilizing anything from styles.module.css, should be cleaned up or removed?


src/consts.ts
```diff
@@ -1,4 +1,12 @@
 const MAILTO_URL = 'mailto:hello@sendchinatownlove.com';
+const emailIcon = require('./images/social-icons/email.png');
+const instagramIcon = require('./images/social-icons/instagram.png');
+const tiktokIcon = require('./images/social-icons/tiktok.png');
+const facebookIcon = require('./images/social-icons/facebook.png');
+const twitterIcon = require('./images/social-icons/twitter.png');
+const linkedinIcon = require('./images/social-icons/linkedin.png');
+const youtubeIcon = require('./images/social-icons/youtube.png');
+const wechatIcon = require('./images/social-icons/weixin.png');
```
wilsonj806, line 9: You should be able to use import syntax on anything that's not a .js/.ts/.tsx file like in the [Hero Banner](https://github.com/sendchinatownlove/sendchinatownlove.github.io/blob/6a63221872407f5a36df84ac6f7eca3081ecec9c/src/components/HeroBanner/HeroBanner.tsx#L3) code

jacoblb64: +1, looks weird and is equivalent, but we generally prefer the `import X from X` syntax instead of the `const ... require()`

src/consts.ts
```diff
@@ -65,8 +73,16 @@ const SquareErrors = {
 };
 
 interface SocialMediaLink {
-  platform: 'facebook' | 'instagram' | 'wechat';
+  platform:
```
jacoblb64, line 76: at this point I'm fine just typing this a string -- or creating a separate enum if we really want to keep locks on this

src/consts.ts
```diff
@@ -106,6 +145,7 @@ const COST_LIMIT_DOLLARS = 10000;
 
 export {
   MAILTO_URL,
+  emailIcon,
```
jacoblb64, line 148: why is this being exported?

jacoblb64 (approved):
Thanks so much for taking care of this! Looks perfect in the screenshots :)

jacoblb64:
[Update Social Media icons in footer](https://trello.com/c/CnFtCzMM/844-update-social-media-icons-in-footer)

src/consts.ts
carolinexsun: Removed this from export

src/consts.ts
carolinexsun: Replaced consts with imports and replaced list of platforms with string type since they're defined in the array later anyway.

jacoblb64 (approved):
Thanks for addressing the comments!

